PR #1, opened 2020-10-17
Function metrics

vlaksi:
I add C# documentation in this form. For someone who is new in this project to know what is some metric and how someone else call her.

image

### DIFF
--- a/RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpMetricCalculator.cs
+++ b/RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpMetricCalculator.cs
@@ -1,7 +1,9 @@
 ﻿using Microsoft.CodeAnalysis.CSharp.Syntax;
 using RepositoryCompiler.CodeModel.CaDETModel;
 using System;
+using System.Collections.Generic;
 using System.Linq;
+using System.Reflection.Metadata;
 
 namespace RepositoryCompiler.CodeModel.CodeParsers.CSharp
 {
@@ -77,8 +79,22 @@ namespace RepositoryCompiler.CodeModel.CodeParsers.CSharp
             return new CaDETMemberMetrics
             {
                 CYCLO = CalculateCyclomaticComplexity(member),
-                LOC = GetLinesOfCode(member.ToString())
+                LOC = GetLinesOfCode(member.ToString()),
+                NOP = GetNumberOfParameters(member),
+                NOLV = GetNumberOfLocalVariables(member)
             };
+        }
+
+        private int GetNumberOfLocalVariables(MemberDeclarationSyntax method)
+        {
+            // TODO: Change this also to real value
+            return 10;
+        }
+
+        private int GetNumberOfParameters(MemberDeclarationSyntax method)
+        {
+            // First because of structure type of 'OfType<ParameterListSyntax>()'
+            return method.DescendantNodes().OfType<ParameterListSyntax>().First().Parameters.Count;
         }
 
         private int GetLinesOfCode(string code)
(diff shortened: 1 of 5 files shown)


vlaksi:
- Added unit tests for NOP & NOLV with manual calculate for metrics

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpMetricCalculator.cs
```diff
@@ -77,10 +79,24 @@ public CaDETMemberMetrics CalculateMemberMetrics(MemberDeclarationSyntax member)
...
+        private int GetNumberOfLocalVariables(MemberDeclarationSyntax method)
+        {
+            // TODO: Change this also to real value
+            return 10;
+        }
+
+        private int GetNumberOfParameters(MemberDeclarationSyntax method)
+        {
+            // First because of structure type of 'OfType<ParameterListSyntax>()'
+            return method.DescendantNodes().OfType<ParameterListSyntax>().First().Parameters.Count;
```
vlaksi, line 97: - I find this way to get number of parameters
- There is some better i guess, so fell free to inform me about it

vlaksi:
## Issue with unit testing

  - I tested ONLY my functions and they work
  - After testing my functions, i just run all other tests and there is issue

image


image

## Problem

Problem is when i include NOP & NOLV in CaDETMemberMetrics class.

image

Without them all other tests works(except my tests).

image


image




@Luburic & @Pufke  have you some fix for this ?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpMetricCalculator.cs
Luburic: This seems alright, but we should make a good suite of test data to see if the results of the calculation are as expected.

The main issue (the reason why other tests are failing) is that you do not handle the edge case with properties that lack parameter lists (even empty ones). Since we are calculating metrics for all types of executable methods (properties, constructors, methods) we need to check them as well.

Try replacing the body with this code:
            var paramLists = method.DescendantNodes().OfType<ParameterListSyntax>().ToList();
            return !paramLists.Any() ? 0 : paramLists.First().Parameters.Count;

Two things to consider:
- What is the situation when we have multiple parameter lists?
- Pro tip: Run all unit tests before committing anything :)

RepositoryCompilerTests/Unit/CaDETCodeModelTests.cs
```diff
@@ -125,5 +126,34 @@ public void Determines_if_is_data_class()
...
+
+        [Fact]
+        public void Calculates_number_of_parameters()
+        {
+            CodeModelBuilder builder = new CodeModelBuilder(LanguageEnum.CSharp);
+
+            List<CaDETClass> classes = builder.BuildCodeModel(_testDataFactory.GetGitAdapterClassText());
+
+            var gitClass = classes.First();
+            gitClass.Methods.Find(method => method.Name.Equals("CheckForNewCommits")).Metrics.NOP.ShouldBe(0);
```
Luburic, line 138: We should create a helper function for finding methods that makes the code more DRY, e.g.

FindMethod(CaDETClass c, String methodName)
{
  return c.Methods.Find(method => method.Name.Equals(methodName));
}

We can then resolve Feature Envy and reduce the parameter count by moving the method to the CaDETClass.

Luburic (changes requested):
You will find the comments inline, including the bugfix. We can discuss the review on Thursday.

Luburic:
[Function metrics](https://trello.com/c/nLtmUBJK/72-function-metrics)

Luburic (commented):
Fixes up to this point look good. When you move the parameter calculation to the Member class and refactor the # of params metric calculation make sure to run all unit tests before the commit.

Luburic (approved):
Lgtm.

PR #2, opened 2020-10-22
Content Repository

vlaksi:
- added class which represent Content
  - added class which represent type of content


vlaksi:
[Base ContentRepository](https://trello.com/c/6n4fvHUS/81-base-contentrepository)

SmartTutor/ContentModel/EducationContent.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+
+namespace SmartTutor.ContentModel
+{
+    public class EducationContent
+    {
+        public int EducationQuality { get; set; }
```
Luburic, line 11: A better name for these two fields would be ContentQuality and ContentDifficulty.

SmartTutor/Controllers/ContentRepositoryService.cs
```diff
@@ -0,0 +1,129 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace SmartTutor.Controllers
+{
+    public class ContentRepositoryService
+    {
+        public ContentRepositoryService() { }
+
+        public EducationContent GetLongParameterListContent()
```
Luburic, line 13: We should think about design here. What does the "ContentRepositoryService" do? Is the name suitable?

As a "service" class, its primary role is to load data, coordinate other, smarter objects, and either save data or help the controller formulate a response. Currently, this class acts as an InMemoryRepository, as it has hardcoded information.

Based on this, a good refactoring would include the renaming of the class to "ContentInMemoryRepository", having it implement the IContentRepository interface to support DIP, and exposing a public method "FindEducationalContent(IssueId), that can (for the purpose of this implementation) call the private methods GetLongParameterListContent and GetLongMethod**Content** (missing in name) based on the detected issues.

Don't forget to move it to a more suitable namespace.

SmartTutor/Controllers/ContentRepositoryService.cs
```diff
@@ -0,0 +1,129 @@
...
+            EducationSnippet snippetFour = new EducationSnippet();
+            EducationSnippet snippetFive = new EducationSnippet();
+
+            snippetOne.SnippetQuality = 4;
+            snippetOne.Tags = new List<Tag>(); snippetOne.Tags.Add(Tag.MustKnow);
+            snippetOne.EducationType = EducationType.ShortText;
+            snippetOne.SnippetDifficulty = 3;
+            snippetOne.Content = "This method have problem with long parameter list";
+            
+
```
Luburic, line 33: In general, there is no need for two blank lines, neither inside nor outside a function.

SmartTutor/ContentModel/EducationSnippet.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace SmartTutor.ContentModel
+{
+    public class EducationSnippet
+    {
+        public int SnippetQuality { get; set; }
+        public int SnippetDifficulty { get; set; }   
+        public EducationType EducationType { get; set; }
```
Luburic, line 12: SnippetType more suitable for both the class and property name?

SmartTutor/Controllers/ContentRepositoryService.cs
```diff
@@ -0,0 +1,129 @@
...
+            snippetTwo.SnippetQuality = 5;
+            snippetTwo.Tags = new List<Tag>(); snippetTwo.Tags.Add(Tag.Interesting);
+            snippetTwo.EducationType = EducationType.LongText;
+            snippetTwo.Content = "The ideal number of arguments for a function is zero (niladic). Next comes one (monadic), followed closely by two (dyadic). Three arguments (triadic) should be avoided where possible. More than three (polyadic) requires very special justification—and then shouldn’t be used anyway.The argument is at a different level of abstraction than the function name and forces you to know a detail that isn’t particularly important.at that point.";
+            snippetTwo.SnippetDifficulty = 3;
+
+            snippetThree.SnippetQuality = 4;
```
Luburic, line 40: It would be useful if you encapsulate each snippet creation into a standalone private function. E.g.:

private EducationalSnippet CreateLongParameterListCodeSnippet() {
            EducationSnippet snippetFour = new EducationSnippet();
            snippetFour.SnippetQuality = 4;
            snippetFour.Tags = new List<Tag>(); snippetFour.Tags.Add(Tag.Interesting);
            snippetFour.EducationType = EducationType.CodeSnippet;
            snippetFour.Content = "Circle makeCircle(double x, double y, double radius);Circle makeCircle(Point center, double radius); ";
            snippetFour.SnippetDifficulty = 5;
            return snippetFour;
}
Just make sure not to call it snippetFour inside this private function :)

This will help us reuse the logic to create different content (e.g., with 2 or 3 snippets and with different ordering).

I would also use the Constructor to instantiate all these objects and place them in a Map, where the issueId is the key.

SmartTutor/Controllers/ContentRepositoryService.cs
```diff
@@ -0,0 +1,129 @@
...
+
+
+            educationContent.EducationSnippets.Add(snippetOne);
+            educationContent.EducationSnippets.Add(snippetTwo);
+            educationContent.EducationSnippets.Add(snippetThree);
+            educationContent.EducationSnippets.Add(snippetFour);
+            educationContent.EducationSnippets.Add(snippetFive);
+
+
+            return educationContent;
```
Luburic, line 126: @vlaksi In general, you should think more about your code. While its ok to have an initial solution look like this, you should not rush for a review without first reviewing the work yourself and looking for refactoring opportunities and the broader design.

Regarding refactoring opportunities, a specific example is the length of these functions and the use of blank lines to clearly separate the multiple things the function is doing.

Regarding the broader design, a specific example is the place of this class in the broader Smart Tutor module. While the class is "by design" meant for temporary use, it still represents a legitimate building block and should be appropriately declared as an InMemoryRepository.

Don't take this the wrong way, it is meant to serve as input for improving your development process.

Luburic (changes requested):
The inline comments give concrete improvements.

We will still need to create a controller and service to enable the Smart Tutor to interact with the other modules, but this can wait while we establish the RepoCompiler to SmellDetector communication.

SmartTutor/Controllers/ContentRepositoryService.cs
vlaksi: First thank you for so much learning opportunities in real examples.

For this class (ContentRepositoryService). i know i rush, my bad. 
I just think how to make it quickly for Milica, and that is bad, i know.

SmartTutor/Controllers/ContentRepositoryService.cs
vlaksi: I understand first part of this, but why we need these objects in a Map ?

vlaksi:
## Infrastructure redesign
</br>

I make new infrastructure instead of one class ( ContentRepositoryService) 

</br>

image

SmartTutor/Controllers/ContentRepositoryService.cs
Luburic: We can use other data structures, but a map (aka C# Dictionary) would give us quick access to educational content for a specific issue. We might also want the value part of the Dictionary to have a List of EduContent. The primary use case is to offer the first from the list (probably the highest quality one), but if the user should click "Give me next lesson" we would retrieve the 2nd element from the list, etc.

Luburic:
> ## Infrastructure redesign
> 
> I make new infrastructure instead of one class ( ContentRepositoryService)
> 
> 
> ###### Click on image for better resolution
> image

@vlaksi I have a couple of comments regarding this design, and look to generalize them to all future designs. There are several issues with the _CreateLongMethods_ and _CreateLongParameterList_ classes:
- First of all, the name implies an action (verb) instead of a noun, which we require of our classes and objects. Since the purpose of these classes is to create some educational snippets, we can use a term such as a Factory (so _LongMethodSnippetFactory_). Furthermore, since the API is the same for both classes, we might opt to define a superclass EducationalSnippetAbstractFactory.
- **However**, we should not expect that all EducationalContent will have these five snippets (video, image, short text, long text, code). Some content might only contain a single long text, a short text and image, or two images. We must ask what purpose do these two classes have - where we can define their responsibility as "Create data for ContentInMemoryRepository". This is a responsibility that can be merged into the InMemoryRepository class if there isn't too much data. Otherwise, we might opt to have one additional class (title something like InMemoryContentFactory). Crucially, we would want a clearly defined API for this - e.g., a public method for each issue that returns one or mote EducationalContent.

The second issue is regarding the controller and service. While we would definitely want to have similar classes to support the Smart Tutor, we should think about better names for these classes. What is the responsibility of the ContentRepositoryService? Currently, it is a middleman between the controller and repository and provides little value. A better name would be ContentService, where we can promote this class to become a coordinator between the different elements of the Smart Tutor - the trainee repository, where we can update information regarding the behavior of the user, the recommender system that calculates suitable educational content based on the trainee and issue, and the content repository, from which we retrieve the information.
For now, we do not need to implement these classes. This will come with @milicasiriski 's work. It is enough to rename the service and make it a more general and useful class, instead of a simple delegator.

Finally, avoid implementing the ContentFileRepo, as we will most likely have a NoSQL database for the content when the time comes (and at that time we will introduce a suitable class).

SmartTutor/Controllers/ContentRepositoryController.cs
```diff
@@ -0,0 +1,23 @@
+﻿using SmartTutor.ContentModel;
+using SmartTutor.Repository;
+using SmartTutor.Service;
+
+namespace SmartTutor.Controllers
+{
+    public class ContentRepositoryController
```
Luburic, line 7: Rename this to "ContentRecommendationController" and we are done 👍

Luburic (approved):

PR #4, opened 2020-10-23
Infrastructure for smelll detector

vlaksi:


vlaksi:
[Infrastructure for smelll detector](https://trello.com/c/qkMOSKhz/84-infrastructure-for-smelll-detector)

SmellDetector/DTO/MetricsDTO.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace SmellDetector.DTO
+{
+    public class MetricsDTO
```
Luburic, line 8: Since DTOs are tightly coupled to controllers and the contract they provide to external services, I would package them with the controller, i.e. in the Controllers folder.

SmellDetector/Controllers/DetectionService.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using SmellDetector.DTO;
+
+public class DetectionService
```
Luburic, line 4: For now we can leave the DetectionService in the Controllers folder, but note that this is not a good location for this service. I imagine you looked to the RepositoryCompiler for inspiration, where the service is misplaced in the Controllers folder as well.

Just make sure you don't follow this poor practice in your PSW project :)

SmellDetector/Controllers/DetectionService.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    public SmellType classCheck(MetricsDTO metrics)
+    {
+        // TODO: Add check for class
+        return SmellType.WITHOUT_BAD_SMELL;
+    }
+
+    public SmellType functionCheck(MetricsDTO metrics)
+    {
+        LongMethod longMethod = new LongMethod();
```
Luburic, line 16: Think about the Open-Closed Principle.

With the current code, there is no way for me to add additional smell detectors without changing existing code. Furthermore, by explicitly separating function-level checks from class-level checks it will be difficult to integrate smell detectors that work at both levels (e.g., mysterious name).

We should establish a system that can be dynamically expanded with multiple detectors. The chain of responsibility comes to mind https://refactoring.guru/design-patterns/chain-of-responsibility . Each detector can work on the CaDETClass, CaDETProject, or even CaDETModel (we can stick to CaDETClass for now). It knows how to get the information it needs from this structure and how to augment the final "SmellDetectionReport" object (e.g., by receiving it from the previous handler as input and returning it after augmentation as output.

Note that your current implementation only says Yes or No based on the supplied metrics. We are looking for more information than that - for the given code what are all the smells, and where are they?

For example, if I pass a class X that has 4 methods A,B,C,D, I want to know from one go that method A is a longMethod, that methods B and C are ok, that method D has a long param list, and that the whole class X has low cohesion and can be split into two.

SmellDetector/SmellDetectionRules/SmellRule.cs
```diff
@@ -0,0 +1,8 @@
+using System;
+using SmellDetector.DTO;
+
+public abstract class SmellRule
```
Luburic, line 4: This should be an interface.

Luburic (changes requested):
One major comment and a few minor ones. Think about the implications that the proposed SmellDetectionReport will have on the Smart Tutor and on the user interface - how will we display multiple smells and suitable educational content for each smell in a user-friendly way. This will be a major challenge that will test our HCI skills :)

SmellDetector/Controllers/DetectionService.cs
vlaksi: Yes, i looked at RepositoryCompiler and i thought we are putting Services in Controllers folder for some reason 😄 

Ofc i will not use this practice in our PSW project 😄

SmellDetector/SmellModel/Issue.cs
```diff
@@ -0,0 +1,15 @@
...
+        public SmellType IssueType { get; set; }
+
+        public string Problem { get; set; }
+
+        public Issue()
+        {
+            IssueType = SmellType.WITHOUT_BAD_SMELL;
+            Problem = "Problem is: ";
+        }
+    }
```
vlaksi, line 14: ## Comment

With this code, I try to reply to what is the smell and where it is.

image

## Example which detects where is the smell

Something like this is the output: **'Problem is Long parameter list in private void AddSomeFeatureExample()'**

I see it is not a generic way and needs more improvement, but for now, I hope is good enough?

image

Luburic: It is enough to have the actual ID (your pair.key) - let the **View** worry about which text will be displayed. Here we are focused on the data - the issues related to a code element.

SmellDetector/SmellModel/PartialSmellDetectionReport.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public void AddIssue(string id, Issue issue)
+        {
+            if (!PartialReport.ContainsKey(id)) PartialReport.Add(id, new List<Issue> { issue });
+            else PartialReport[id].Add(issue);
+        }
+
+        public void AddIssues(string id, List<Issue> issues)
+        {
+            if (!PartialReport.ContainsKey(id)) PartialReport.Add(id, issues);
```
Luburic, line 22: You have duplicate logic here. You can transform AddIssue:
`public void AddIssue(string id, Issue issue)
{
  AddIssues(id, new List<Issue> { issue });
}`

Or go the other way around and have AddIssues call AddIssue.

Also lets rename the id to be codeItemId and make it more explicit.

Luburic (changes requested):
Inline comments.

SmellDetector/SmellModel/SmellDetectionReport.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class SmellDetectionReport
+    {
+        public Dictionary<string,List<Issue>> Report { get; set; }
+
+        public SmellDetectionReport()
+        {
+            Report = new Dictionary<string, List<Issue>>();
+        }
+
+        public void AddPartialReport(PartialSmellDetectionReport partialReport)
```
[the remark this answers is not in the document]
vlaksi: Thanks.

Luburic (changes requested):
We should work on the "sloppiness" factor a bit. I think that we are rushing towards the next challenge without properly resolving the previous one.

A few tips:
- At a minimum, we must always ask "does this work" before considering finishing a piece of work. When working on complex systems our only options are console applications to instantiate and run the part of the system that we are building or automated tests - which bring many other benefits.
- The next step is to think about edge cases and null - where is the input to the system, under which circumstances can we have null or invalid values, etc. This does not mean to apply null checks or validation at every possible location - just think about where things might go wrong (ideally only at the module's boundary).
- Refactoring and thinking about places for improvement.

SmellDetector/Detectors/RuleEngines/LongMethodRuleEngine.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace SmellDetector.Detectors.RuleEngines
+{
+    public class LongMethodRuleEngine : IDetector
+    {
+
+        public PartialSmellDetectionReport FindIssues(CaDETClassDTO caDetClassDto)
+        {
+            PartialSmellDetectionReport partialReport = new PartialSmellDetectionReport();
+
+            foreach (var pair in caDetClassDto.CaDETClass)
```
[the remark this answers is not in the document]
vlaksi: This is an awesome algorithm. From now I have an idea of how to solve problems with a not 'good name'. Thanks!

SmellDetector/SmellModel/CaDETClassDTO.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.Generic;
+using SmellDetector.Controllers;
+
+namespace SmellDetector.SmellModel
+{
+    public class CaDETClassDTO
+    {
+        public Dictionary<string, MetricsDTO> IdentifierAnalyses { get; set; }
```
Luburic, line 8: Still a pretty mysterious name "Analize identifikatora". Metric calculations aren't really analyses, and its unclear what identifier refers to.

We need to think about what we are receiving from the RepoCompiler https://github.com/Clean-CaDET/platform-documentation where the answer is metrics and features for various code items. Since we currently have only metrics, we can rename this field to something like CodeItemMetrics.

SmellDetector/SmellModel/Reports/PartialSmellDetectionReport.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections.Generic;
+
+namespace SmellDetector.SmellModel.Reports
+{
+    public class PartialSmellDetectionReport
+    {
+        public Dictionary<string, List<Issue>> PartialReport { get; set; }
```
Luburic, line 7: A field called PartialReport in a class that is essentially called PartialReport is not a good name 🙂 

Following the previous renaming, we can call this CodeItemIssues.

SmellDetector/SmellModel/Reports/SmellDetectionReport.cs
```diff
@@ -0,0 +1,32 @@
...
+        public Dictionary<string,List<Issue>> Report { get; set; }
+
+        public SmellDetectionReport()
+        {
+            Report = new Dictionary<string, List<Issue>>();
+        }
+
+        public void AddPartialReport(PartialSmellDetectionReport partialReport)
+        {
+            foreach (var pair in partialReport.PartialReport)
```
Luburic, line 16: foreach(var codeItemIssue in partialReport.codeItemIssues)

Luburic (changes requested):
Few renames and we are done.

Luburic (approved):
Well done 🙂

PR #5, opened 2020-10-28
Refactored code parser

Luburic:
This is a major change that should break some of your code from the active PRs.

The additions include:
- Defining a CaDETMemberParamter class that encapsulates the name and type.
- Extracting a CaDETField as a standalone type from the current CaDETMember.
- Adding the CSharpCaDETMemberTransformer to split the logic of the CSharpCodeParser.

Please review the code, leave any comments for mysterious names or stuff that is hard for you to understand so that I can enhance it further.

### DIFF
--- a/RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
+++ b/RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
@@ -7,14 +7,14 @@ namespace RepositoryCompiler.CodeModel.CaDETModel.CodeItems
 {
     public class CaDETClass
     {
-        public string Name { get; set; }
-        public string FullName { get; set; }
-        public string SourceCode { get; set; }
-        public CaDETClass Parent { get; set; }
-        public List<CaDETModifier> Modifiers { get; set; }
-        public List<CaDETMember> Members { get; set; }
-        public List<CaDETField> Fields { get; set; }
-        public CaDETClassMetrics Metrics { get; set; }
+        public string Name { get; internal set; }
+        public string FullName { get; internal set; }
+        public string SourceCode { get; internal set; }
+        public CaDETClass Parent { get; internal set; }
+        public List<CaDETModifier> Modifiers { get; internal set; }
+        public List<CaDETMember> Members { get; internal set; }
+        public List<CaDETField> Fields { get; internal set; }
+        public CaDETClassMetrics Metrics { get; internal set; }
 
         public CaDETMember FindMember(string name)
         {
@@ -28,7 +28,7 @@ namespace RepositoryCompiler.CodeModel.CaDETModel.CodeItems
 
         public bool IsDataClass()
         {
-            double numOfAccessors = Members.Count(m => m.IsSimpleAccessor());
+            double numOfAccessors = Members.Count(m => m.IsFieldDefiningAccessor());
             double numOfConstructors = Members.Count(m => m.Type.Equals(CaDETMemberType.Constructor));
             double numOfObjectOverrides = CountToStringEqualsHashCode();
             //TODO: Create a more elegant solution for thresholds.
--- a/RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
+++ b/RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
@@ -12,29 +12,34 @@ namespace RepositoryCompiler.CodeModel.CaDETModel.CodeItems
         public string SourceCode { get; set; }
         public CaDETClass Parent { get; set; }
         public List<CaDETModifier> Modifiers { get; set; }
-        public List<CaDETMember> Methods { get; set; }
-        public List<CaDETMember> Fields { get; set; }
+        public List<CaDETMember> Members { get; set; }
+        public List<CaDETField> Fields { get; set; }
         public CaDETClassMetrics Metrics { get; set; }
 
-        public CaDETMember FindMethod(string name)
+        public CaDETMember FindMember(string name)
         {
-            return Methods.Find(method => method.Name.Equals(name));
+            return Members.Find(method => method.Name.Equals(name));
+        }
+
+        public CaDETField FindField(string name)
+        {
+            return Fields.Find(field => field.Name.Equals(name));
         }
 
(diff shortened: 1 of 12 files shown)


RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
```diff
@@ -7,34 +7,44 @@ namespace RepositoryCompiler.CodeModel.CaDETModel.CodeItems
...
-        public List<CaDETModifier> Modifiers { get; set; }
-        public List<CaDETMember> Methods { get; set; }
-        public List<CaDETMember> Fields { get; set; }
-        public CaDETClassMetrics Metrics { get; set; }
+        public string Name { get; internal set; }
+        public string FullName { get; internal set; }
+        public string SourceCode { get; internal set; }
+        public CaDETClass Parent { get; internal set; }
+        public List<CaDETModifier> Modifiers { get; internal set; }
+        public List<CaDETMember> Members { get; internal set; }
```
Pufke, line 15: This is a much better name. I like it. 💯
There should be no confusion now.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETField.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections.Generic;
+
+namespace RepositoryCompiler.CodeModel.CaDETModel.CodeItems
+{
+    public class CaDETField
+    {
+        public string Name { get; internal set; }
+        public List<CaDETModifier> Modifiers { get; internal set; }
```
Pufke, line 8: What does it represent field "Modifiers" ?

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETMemberType.cs
```diff
@@ -4,7 +4,6 @@ public enum CaDETMemberType
     {
         Method,
         Property,
-        Constructor,
-        Field
```
Pufke, line 8: Field is no longer belongs to Member ? This is reason why you make separate class CaDETField.cs ?

Luburic: Yes. In the previous version, all instances of CaDETMember that had the type field had most of their fields equal to null (as fields don't have accessedAccessors, invokedMethods and metrics, for example).

While this somewhat complicates the rest of the code (as we can't treat fields and other executable code (methods, properties, constructors) the same way, it should reduce confusion and bugs in the long run.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETMember.cs
```diff
@@ -1,43 +1,62 @@
...
     public class CaDETMember
     {
-        public string Name { get; set; }
-        public CaDETMemberType Type { get; set; }
-        public string SourceCode { get; set; }
-        public CaDETClass Parent { get; set; }
-        public List<string> Params { get; set; }
-        public List<CaDETModifier> Modifiers { get; set; }
-        public ISet<CaDETMember> InvokedMethods { get; set; }
-        public ISet<CaDETMember> AccessedFieldsAndAccessors { get; set; }
```
Pufke, line 16: Separating Fields and Accessors is good idea.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETField.cs
Luburic: Modifiers are keywords that change the behavior/meaning of a function, field, or class. These include "static", "abstract", and visibility modifiers such as private, public, internal...
Currently we do not utilize them in our smell detectors, but it might be interesting to see if a class has for example 10+ public methods and few private methods - signaling that it might have too many responsibilities.

Pufke (approved):

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
vlaksi: ### Comment

Member is good for C# but, is it good for other languages ? For now this is good, and i agree but definitely we will need to refactor this to make it more suitable for other languages.

### Proposal

When we start to work on Java and others, make brainstorming about this and resolve this.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETMember.cs
```diff
@@ -1,43 +1,62 @@
...
+                {
+                    sb.Append(Params[i].Type);
+                    if (i < Params.Count - 1) sb.Append(", ");
+                }
+                sb.Append(")");
+            }
+
+            return sb.ToString();
+        }
+
```
vlaksi, line 44: ### Comment

I understand what is sb, but i need few seconds to understand.

### Proposal

Instead of name sb, more readable is maybe signature:

```sh
        public string GetSignature()
        {
            var signature = new StringBuilder();
            if (Parent != null) signature.Append(Parent.FullName).Append(".");
            signature.Append(Name);
            if (Params != null)
            {
                signature.Append("(");
                for (var i = 0; i < Params.Count; i++)
                {
                    signature.Append(Params[i].Type);
                    if (i < Params.Count - 1) signature.Append(", ");
                }
                signature.Append(")");
            }

            return signature.ToString();
        }
```

vlaksi (changes requested):
The inline comments give concrete improvements.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
Luburic: Agreed. We also might decompose Members into accessors, properties and constructors in time.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETMember.cs
Luburic: "sb" is often used as a name for an instance of the StringBuilder class. To make it more explicit I'll name it signatureBuilder.

PR #6, opened 2020-10-29
ATFD Metric

Pufke:
Definition of ATFD:
    • Access to Foreign Data (ATFD) represents the
      number of attributes of foreign classes accessed
      directly or by using accessor methods. 

Reference: DOI: 10.1109/ESEM.2009.5314231

### DIFF
--- a/RepositoryCompiler/CodeModel/CaDETModel/Metrics/CaDETClassMetrics.cs
+++ b/RepositoryCompiler/CodeModel/CaDETModel/Metrics/CaDETClassMetrics.cs
@@ -7,6 +7,10 @@
         public int NMD { get; set; }
         public int NAD { get; set; }
         public int WMC { get; set; }
+        /// <summary>
+        /// ATFD: Access To Foreign Data
+        /// DOI: 10.1109/ESEM.2009.5314231
+        /// </summary>
         public int ATFD { get; set; }
     }
 }
--- a/RepositoryCompiler/CodeModel/CaDETModel/Metrics/CaDETClassMetrics.cs
+++ b/RepositoryCompiler/CodeModel/CaDETModel/Metrics/CaDETClassMetrics.cs
@@ -7,14 +7,7 @@
         public int NMD { get; set; }
         public int NAD { get; set; }
         public int WMC { get; set; }
-        /// <summary>
-        /// ATFD: Access To Foreign Data
-        /// DOI: 10.1109/ESEM.2009.5314231
-        /// </summary>
         public int ATFD { get; set; }
-        /// TCC - Tight Class Cohesion
-        /// DOI: 10.1145/223427.211856
-        /// </summary>
         public double? TCC { get; set; }
     }
 }
--- a/RepositoryCompiler/CodeModel/CaDETModel/Metrics/CaDETClassMetrics.cs
+++ b/RepositoryCompiler/CodeModel/CaDETModel/Metrics/CaDETClassMetrics.cs
@@ -7,5 +7,6 @@
         public int NMD { get; set; }
         public int NAD { get; set; }
         public int WMC { get; set; }
+        public int ATFD { get; set; }
     }
 }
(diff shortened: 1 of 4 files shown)


RepositoryCompiler/CodeModel/CaDETModel/Metrics/CaDETClassMetrics.cs
```diff
@@ -8,6 +8,10 @@ public class CaDETClassMetrics
         public int NAD { get; set; }
         public int WMC { get; set; }
         /// <summary>
+        /// ATFD: Access To Foreign Data
+        /// DOI: 10.1109/ESEM.2009.5314231
+        /// </summary>
+        public int ATFD { get; set; }
```
Luburic, line 14: I think I mislead you with the DOI, and placing it in this class.

We would want the DOI to be in the summary comment of the actual function doing the calculation, as the algorithm is what is closely tied to the DOI, not necessarily the original definition of the metric. You should still leave the full name in the summary (and maybe even add alternative names from that overview paper that we keep referencing).

You can take a look at Vladislav's MemberMetrics and functions in the Calculator to see how it should look.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpMetricCalculator.cs
```diff
@@ -20,10 +20,23 @@ public CaDETClassMetrics CalculateClassMetrics(CaDETClass parsedClass)
...
             };
         }
 
+        private int GetAccessToForeignData(CaDETClass parsedClass)
+        {
+            int numberOfAccessToForeignFieldsAndAccessors = 0;
+
+            foreach (var member in parsedClass.Members)
+            {
+                numberOfAccessToForeignFieldsAndAccessors += (member.AccessedFields.Count - member.GetAccessedOwnFields().Count) + (member.AccessedAccessors.Count - member.GetAccessedOwnAccessors().Count);
```
Luburic, line 34: I was of the opinion that we concluded that there were 3+ references that state that ATFD counts the number of external classes that we are accessing either through a field or accessor. Here we are counting the actual number of foreign fields and accessors accessed.

Is this correct? I remember that the thresholds for ATFD were between 2 and 5. Could you list the few papers (e.g., by name) based on which we are going with this calculation?

RepositoryCompilerTests/Unit/CodeModelTestDataFactory.cs
```diff
@@ -162,6 +162,138 @@ private Repository GetRepository()
...
+            using System.Collections.Generic;
+            namespace NDCApp.Model.Data
+            {
+                  public class Class3
+                  {
+                   
+                    public double m1(){
+                        Class4 class4 = new Class4();
+                       
+                        class4.Hours = 23;
```
Luburic, line 206: I like the test data suite for the ATFD, but I think it's missing an important case when we have multiple methods in a class, as well as when we are referencing multiple classes and their fields. Such a case is especially important if we change the ATFD calculation.

Maybe change classes 3 and 4 to reference some subset of the other 6 classes in some way, as I believe classes 5 and 6 cover the test case that you currently have with 3 and 4.

Luburic (changes requested):
Give info regarding the calculation formula and enhance the test data.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpMetricCalculator.cs
Luburic: @Pufke In the event that we are indeed tracking the different classes and not the fields, we should take inheritance in to account.
Namely, lets examine the following situation:
- Class A defines public field/accessor X, and class B inherits A and defines public field/accessor Y.
- Class C references B and accesses both Y and X.
- ATFD for C should be one.

We can write a test for this, in the event that ATFD will indeed count distinct classes.

Pufke: Yea, at the beginning of the research we had that opinion.

But after researching, we concluded the following: Access To Foreign Data (ATFD) is the number of
attributes in foreign classes accessed by class. 

 References from the following paper:
 Bad smell detection using quality metrics and refactoring
 opportunities 
 DOI: 10.1002/smr.2255

 References: 

-  Macia I, Garcia J, Popescu D, Garcia A, Medvidovic N, von Staa A. Are automatically-detected code 
   anomalies relevant to architectural modularity
   DOI: 10.1145/2162049.2162069
   Screenshot from paper: 
image

-   The Evolution and Impact of Code Smells: A Case Study of Two Open Source Systems 
    DOI: 10.1109/ESEM.2009.5314231
    Screenshot from paper:
image

PR #7, opened 2020-10-29
God Class Education Material

Pufke:


Luburic (commented):
This is enough for our proof of concept, as we are focusing on the whole end-to-end pipeline and the infrastructure.

If you have any more materials related to the Large class code smell in your notes, it would be helpful to put them here. If not, someone else will work on creating better educational content (this is something we need to do in general, as our current content is pretty simplistic).

Pufke:
In my notes I have content that is scientifically technologically grounded. Content already listed in PR (Rule engine for God Class) link: https://github.com/Clean-CaDET/platform-repository-compiler/pull/9 

I'm not sure how useful this content is for educational purposes?

PR #11, opened 2021-03-09
CC class cohesion metric and tests

balsa-sarenac:
## Introducing CC class cohesion metric

DOI: [10.14257/ijmue.2014.9.6.35](https://www.researchgate.net/deref/http%3A%2F%2Fdx.doi.org%2F10.14257%2Fijmue.2014.9.6.35)


RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -123,5 +124,43 @@ private int GetNumberOfMethodsDeclared(CaDETClass parsedClass)
...
+        /// CC - Class cohesion
+        /// DOI: 10.14257/ijmue.2014.9.6.35
+        /// </summary>
+        private double? GetCCClassCohesion(CaDETClass parsedClass)
+        {
+            IEnumerable<CaDETMember> memberMethods = parsedClass.Members.Where(m => m.Type != CaDETMemberType.Property);
+            IEnumerable<CaDETMember> memberProperties = parsedClass.Members.Where(m => m.Type == CaDETMemberType.Property);
+            double fieldsCount = parsedClass.Fields.Count + memberProperties.Count();
+
+            if (fieldsCount == 0 || memberMethods.Count() == 0)
```
Luburic, line 138: What happens if fieldCount is zero, but we have field defining properties (e.g., string SomeProp {get;set;})?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -123,5 +124,43 @@ private int GetNumberOfMethodsDeclared(CaDETClass parsedClass)
...
+            {
+                return 0;
+            }
+
+            double cv = CalculateCohesionForEveryVariable(parsedClass, memberMethods, memberProperties);
+
+            return Math.Round(cv / fieldsCount, 2);
+        }
+
+        private static double CalculateCohesionForEveryVariable(CaDETClass parsedClass, IEnumerable<CaDETMember> memberMethods, IEnumerable<CaDETMember> memberProperties)
```
Luburic, line 148: Variable is an odd name. Since we are looking for a common name for fields and properties, I googled this article https://stackoverflow.com/questions/26671085/is-there-a-common-word-in-use-for-fields-and-properties.

Since the term "Member" is used to encapsulate methods, fields, and properties, a comment suggested we call it "DataMember" and I like this.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -123,5 +124,43 @@ private int GetNumberOfMethodsDeclared(CaDETClass parsedClass)
...
+        {
+            double cc = 0;
+
+            foreach (var field in parsedClass.Fields)
+            {
+                double methodsInvokingGivenField = memberMethods.Count(m => m.AccessedFields.Contains(field));
```
Luburic, line 154: The CaDETMember offers the following methods: https://github.com/Clean-CaDET/platform/blob/c3fa836fd7918b5b3c392525c4353c0d7a9482f7/RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETMember.cs#L50-L76

Would something like this (memberMethods.Sum(m => m.GetAccessedOwnFields().Count) + memberMethods.Sum(m => m.GetAccessedOwnAccessors().Count)) / memberMethods.Count give a good result?

If yes, we would also reduce the params of this function to 1 which is a plus.

Luburic (changes requested):
Comments inline.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
balsa-sarenac: fieldsCount is number of all fields and members that are properties - defined in 136. Maybe renaming the variable to "DataMembersCount" like you suggested bellow would make this more clear?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -123,5 +124,43 @@ private int GetNumberOfMethodsDeclared(CaDETClass parsedClass)
...
         }
+
+        /// <summary>
+        /// CC - Class cohesion
+        /// DOI: 10.14257/ijmue.2014.9.6.35
+        /// </summary>
+        private double? GetCCClassCohesion(CaDETClass parsedClass)
+        {
+            IEnumerable<CaDETMember> memberMethods = parsedClass.Members.Where(m => m.Type != CaDETMemberType.Property);
+            IEnumerable<CaDETMember> memberProperties = parsedClass.Members.Where(m => m.Type == CaDETMemberType.Property);
```
Luburic, line 135: IsFieldDefiningAccessor

balsa-sarenac:
This metric is the same as LCOM (which is already implemented) therefore redundant.

PR #12, opened 2021-03-11
Feature/challenges and gamification

Ana00000:
- All activities worth points are types of EducationActivity. They have a life cycle and some sort of content (project, text, image..).
- The idea behind challenges:
 --> Player gets the code for the challenge through cloning. He, then, opens cloned project manually from the folder. Changed project is saved in the same place. Not yet implemented methods will check metrics for new state of project.

### DIFF
--- a/SmartTutor/Controllers/ChallengeController.cs
+++ b/SmartTutor/Controllers/ChallengeController.cs
@@ -1,0 +1,22 @@
+﻿using SmartTutor.ActiveEducationModel;
+using SmartTutor.Repository.ChallengeRepository;
+using SmartTutor.Service;
+
+namespace SmartTutor.Controllers
+{
+    public class ChallengeController
+    {
+        public ChallengeService ChallengeService;
+
+        public ChallengeController()
+        {
+            ChallengeService = new ChallengeService(new ChallengeRepository());
+        }
+
+        public void StartChallenge(SmellType issue, int indexOfProject, Player player)
+        {
+            ChallengeService.StartChallenge(issue, indexOfProject, player);
+        }
+
+    }
+}
--- a/SmartTutor/Controllers/ChallengeController.cs
+++ b/SmartTutor/Controllers/ChallengeController.cs
@@ -17,6 +17,5 @@ namespace SmartTutor.Controllers
         {
             ChallengeService.StartChallenge(issue, indexOfProject, player);
         }
-
     }
 }
--- a/SmartTutor/Controllers/ChallengeProjectController.cs
+++ b/SmartTutor/Controllers/ChallengeProjectController.cs
@@ -1,0 +1,22 @@
+﻿using SmartTutor.ActiveEducationModel;
+using SmartTutor.Repository.ChallengeProjectRepository;
+using SmartTutor.Service;
+
+namespace SmartTutor.Controllers
+{
+    public class ChallengeProjectController
+    {
+
+        public ChallengeProjectService ChallengeProjectService;
+
+        public ChallengeProjectController()
+        {
+            ChallengeProjectService = new ChallengeProjectService(new ChallengeProjectRepository());
+        }
+
+        public ChallengeProject FindProjectForIssue(SmellType issue, int indexOfProject)
+        {
+            return ChallengeProjectService.FindProjectForIssue(issue, indexOfProject);
+        }
+    }
+}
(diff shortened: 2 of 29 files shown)


SmartTutor/Controllers/ChallengeController.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class ChallengeController
+    {
+        public ChallengeService ChallengeService;
+
+        public ChallengeController()
+        {
+            ChallengeService = new ChallengeService(new ChallengeRepository());
+        }
+
+        public void StartChallenge(SmellType issue, int indexOfProject, Player player)
```
Luburic, line 16: This controller looks like it will interface with the IDE plugin.

Since we will look at challenges as learning objects that are part of the "Procedural knowledge" (PK) knowledge node (KN), the start challenge would load the KN with the given ID (so the plugin would also need a function to get Lectures and then get PK KNs for the selected lecture - this can be a single function that sends KN DTOs). **Note that challenges are related to Lectures, not smells**.

SmartTutor/Controllers/ChallengeProjectController.cs
```diff
@@ -0,0 +1,22 @@
+﻿using SmartTutor.ActiveEducationModel;
+using SmartTutor.Repository.ChallengeProjectRepository;
+using SmartTutor.Service;
+
+namespace SmartTutor.Controllers
+{
+    public class ChallengeProjectController
```
Luburic, line 7: We should refactor this to reflect the new design. I think our new view simplifies the overall design a lot, so examine the classes and think about the objects that will be instantiated (and how they will interact) when a request from the IDE arrives.

SmartTutor/Repository/ActiveEducationRepository/ActivityFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+        private EducationTraining CreateLongMethodTraining()
+        {
+            EducationTraining longMethodTraining = new EducationTraining
+            {
+                Start = DateTime.Now,
+                End = DateTime.Now,
+                Status = ActivityStatus.Unlocked,
+                Player = new Player
+                {
+                    Credentials = new UsernamePasswordCredentials { Username = "MIRKOMM", Password = "963mmirkom369" },
```
Luburic, line 56: As stated above, we will remove the credentials from the Player. As a side note, we should rename Player to Trainee.

Btw, are these real credentials? You really don't want to store passwords in your code - git will remember them while this repository exists (which is publicly available). Because of this, there are many tools for searching GitHub to locate and steal credentials from the code.

SmartTutor/Repository/ActiveEducationRepository/ActivityRepository.cs
```diff
@@ -0,0 +1,22 @@
+﻿using SmartTutor.ActiveEducationModel;
+using SmartTutor.Repository.ActiveEducationRepository;
+using System.Collections.Generic;
+
+namespace SmartTutor.Repository
+{
+    public class ActivityRepository : IActiveEducationRepository
```
Luburic, line 7: ActivityInMemoryRepository would be more suitable, although I would change the "Activity" part based on the previous comments.

Activity is a really generic word that should be avoided (especially as a primary identifier of a concept, like its used here).

SmartTutor/Repository/ChallengeProjectRepository/IChallengeProjectRepository.cs
```diff
@@ -0,0 +1,9 @@
+﻿using SmartTutor.ActiveEducationModel;
+
+namespace SmartTutor.Repository.ChallengeProjectRepository
+{
+    public interface IChallengeProjectRepository
```
Luburic, line 5: Since we'll be looking at challenges through the LO view, we can merge this with that part of the content model.

Luburic (changes requested):
I've placed a bunch of inline comments that you should examine.

I also have a few general comments that should help us guide future development.

A PR that changes many things (in this case 29 files) is a dangerous PR, especially when paired with experimental and unknown features. It can result in a lot of wasted time if it turns out that the experiment failed or that we were going in the wrong direction from the start. In general, we should always look to define the minimum functionality required to test an idea - to decompose a user story to its essence, so that we may check if we understand the essence before putting on additional features and implementing all the details. Here we introduce challenges for multiple smelltypes, players, scoring, and modify the RepositoryCompiler to interact with Git in a new way. I would argue that the most basic challenge that has no smell type, no git interaction, no knowledge of players or scores, just an end state would help us understand the essence of a challenge.

Now, I don't think this was wasted work in general. It helped us develop a better understanding of what we want to build and that's the most valuable thing in development (the worst thing is building the wrong product). Hopefully, the inline comments will also help you write better code. 

That being said, moving forward I would like to see smaller PRs that are focused on a specific problem. Always keep in mind the question "what is the smallest thing I can build to test our understanding of a feature". Also look to first implement the most important and unknown or difficult features (the high risk stuff), as this can dramatically affect our design. The bells and whistles (easy to understand stuff that requires more manual work than thinking) come at the end.

PR #13, opened 2021-03-19
Implemented God Class code smell detection based on heuristics

katarina-glorija:
Started new model for code smell detection.

### DIFF
--- a/SmellDetector/Controllers/OrCriteria.cs
+++ b/SmellDetector/Controllers/OrCriteria.cs
@@ -1,0 +1,19 @@
+﻿using System;
+namespace SmellDetector.Controllers
+{
+    public class OrCriteria:Criteria
+    {
+        private Criteria firstCriteria;
+        private Criteria secondCriteria;
+        public OrCriteria(Criteria firstCriteria, Criteria secondCriteria)
+        {
+            this.firstCriteria = firstCriteria;
+            this.secondCriteria = secondCriteria;
+        }
+
+        public bool meetCriteria()
+        {
+            return firstCriteria.meetCriteria() || secondCriteria.meetCriteria();
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


SmellDetector/Controllers/OrCriteria.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+namespace SmellDetector.Controllers
```
Luburic, line 2: These classes should be in the Detectors-RuleEngines namespace.

SmellDetector/Enumerations/OperationEnum.cs
```diff
@@ -0,0 +1,8 @@
+﻿public enum OperationEnum
```
Luburic, line 1: Enums should also be a part of the namespace. This file should be moved alongside the Criteria.

SmellDetector/Services/DetectionService.cs
```diff
@@ -21,7 +21,8 @@ private void LoadDetectors()
             {
                 new LongMethodRuleEngine(),
                 new LongParameterListRuleEngine(),
-                new GodClassRuleEngine(),
+                new GodClassRuleEngine(47, 2, 0.33),
```
Luburic, line 24: Using the new design, this Service will only instantiate the ClassMetricsRuleEngine which will instantiate the rules that can have default values for their thresholds.

Luburic (changes requested):
We need to change a lot of things. First, I recommend installing Visual Studio and the Resharper plugin (both are free with uns email). Their cleanup tools should help maintain cleaner code.

Next, we should redesign our classes to following the design given below:
image
Note: This is a partial design and was entirely written in PowerPoint so you will probably need to adjust a few things.

Also examine the inline comments along the way.

PR #15, opened 2021-03-23
Feature/method and class metrics

simsimkic:


RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
             }
             return count;
         }
+
+        private int CountTryCatchBlocks(MemberDeclarationSyntax method)
```
Luburic, line 141: Based on this code, the metric will add +1 for a try block with a single catch and +1 for a try block with 10 catch statements. Is this intended?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+
+        private int CountReturnStatements(MemberDeclarationSyntax method)
+        {
+            return method.DescendantNodes().OfType<ReturnStatementSyntax>().Count();
+        }
+
+        private int CountComparisonOperators(MemberDeclarationSyntax method)
+        {
+            return method.DescendantNodes().OfType<BinaryExpressionSyntax>()
+                .Count(n => n.IsKind(SyntaxKind.EqualsExpression) ||
```
Luburic, line 163: Would this also count https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator ?

If I'm not mistaking, this would map to "CoalesceExpression" but you should test it (by writing a test/expanding test data).

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+                            n.IsKind(SyntaxKind.GreaterThanExpression) ||
+                            n.IsKind(SyntaxKind.GreaterThanOrEqualExpression));
+        }
+
+        private int CountMethodInvocations(MemberDeclarationSyntax method)
+        {
+            return method.DescendantNodes().OfType<InvocationExpressionSyntax>().Count();
+        }
+
+        private int CountUniqueMethodInvocations(MemberDeclarationSyntax method)
```
Luburic, line 176: https://github.com/Clean-CaDET/platform/blob/master/RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETMember.cs#L16

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+        {
+            return method.DescendantNodes().OfType<LiteralExpressionSyntax>().Count(n => n.IsKind(SyntaxKind.NumericLiteralExpression));
+        }
+
+        private int CountNumberOfStringLiterals(MemberDeclarationSyntax method)
+        {
+            return method.DescendantNodes().OfType<LiteralExpressionSyntax>().Count(n => n.IsKind(SyntaxKind.StringLiteralExpression));
+        }
+
+        private int CountNumberOfMathOperations(MemberDeclarationSyntax method)
```
Luburic, line 205: This will also increment for "Pera " + "Peric". Is that what we want?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+
+        private int CountBinaryExpressions(MemberDeclarationSyntax method)
+        {
+            return method.DescendantNodes().OfType<BinaryExpressionSyntax>()
+                .Count(n => n.IsKind(SyntaxKind.AddExpression) ||
+                            n.IsKind(SyntaxKind.SubtractExpression) ||
+                            n.IsKind(SyntaxKind.MultiplyExpression) ||
+                            n.IsKind(SyntaxKind.DivideExpression) ||
+                            n.IsKind(SyntaxKind.ModuloExpression) ||
+                            n.IsKind(SyntaxKind.LeftShiftExpression) ||
```
Luburic, line 220: The shift operators are counted as math operations?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+        }
+
+        private int CountNumberOfLambdaExpressions(MemberDeclarationSyntax method)
+        {
+            return method.DescendantNodes().OfType<LambdaExpressionSyntax>().Count();
+        }
+
+        private int CountMaxNestedBlocks(MemberDeclarationSyntax method)
+        {
+            if (!method.Kind().Equals(SyntaxKind.MethodDeclaration))
```
Luburic, line 250: Write such return blocks as one line, i.e. if (!method.Kind().Equals(SyntaxKind.MethodDeclaration)) return 0;

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+            return blocksAncestors.Max();
+        }
+
+        private int CountNumberOfUniqueWords(MemberDeclarationSyntax method)
+        {
+            if (!method.Kind().Equals(SyntaxKind.MethodDeclaration))
+            {
+                return 0;
+            }
+            BaseMethodDeclarationSyntax baseMethod = (BaseMethodDeclarationSyntax)method;
```
Luburic, line 276: Do you need code from 272 to 277? Look how the LOC method is used - member.ToString().

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+            return method.DescendantNodes().OfType<LambdaExpressionSyntax>().Count();
+        }
+
+        private int CountMaxNestedBlocks(MemberDeclarationSyntax method)
+        {
+            if (!method.Kind().Equals(SyntaxKind.MethodDeclaration))
+            {
+                return 0;
+            }
+            BaseMethodDeclarationSyntax baseMethod = (BaseMethodDeclarationSyntax)method;
```
Luburic, line 254: Why is this cast important? Can we not extract the descendant nodes from the member, or will this include ancestors from the class and namespace level?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+                .ToList();
+            return RemoveMultiLineComments(lines);
+        }
+
+        private List<string> RemoveMultiLineComments(List<string> lines)
+        {
+            List<string> linesToRemove = new List<string>();
+            for (int i = 0; i < lines.Count(); i++)
+            {
+                if (lines[i].StartsWith("/*"))
```
Luburic, line 300: A multiline comment does not need to be at the start of a line.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+                {
+                    linesToRemove.Add(lines[i]);
+                    int j = 1;
+                    while (!lines[i + j].EndsWith("*/"))
```
Luburic, line 304: Nor does it need to end at the end of the line :)

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+
+        private int CountNumberOfUniqueWords(MemberDeclarationSyntax method)
+        {
+            if (!method.Kind().Equals(SyntaxKind.MethodDeclaration))
+            {
+                return 0;
+            }
+            BaseMethodDeclarationSyntax baseMethod = (BaseMethodDeclarationSyntax)method;
+            string methodBody = baseMethod.Body.ToString();
+            List<string> lines = GetLinesWithoutComments(methodBody);
```
Luburic, line 278: Are you sure that words in comment should not be included?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+            List<string> individualWords = BreakWords(words);
+            individualWords = FindAdditionalWords(individualWords);
+            return CleanWordsFromSymbols(individualWords).Distinct().ToList().Count();
+        }
+
+        private List<string> GetLinesWithoutComments(string methodBody)
+        {
+            List<string> lines = methodBody.Split("\n")
+                .Select(line => line.Trim())
+                .Where(line => !line.StartsWith("//"))
```
Luburic, line 290: Single line comments do not need to start the line (e.g., int a = 5; //This sets a to 5.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+                    while (!lines[i + j].EndsWith("*/"))
+                    {
+                        linesToRemove.Add(lines[i + j]);
+                        j++;
+                    }
+                    linesToRemove.Add(lines[i + j]);
+                }
+            }
+            lines.RemoveAll(line => linesToRemove.Contains(line));
+            return lines;
```
Luburic, line 313: The code to solve this problem is located a few hundred lines above: https://github.com/Clean-CaDET/platform/blob/master/RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs#L31-L37 which you can extract into a private method that will be used by both calculations.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+            words = FilterWords(words);
+            List<string> individualWords = BreakWords(words);
+            individualWords = FindAdditionalWords(individualWords);
+            return CleanWordsFromSymbols(individualWords).Distinct().ToList().Count();
+        }
+
+        private List<string> GetLinesWithoutComments(string methodBody)
+        {
+            List<string> lines = methodBody.Split("\n")
+                .Select(line => line.Trim())
```
Luburic, line 289: Once you utilize the comment extracting code I comment on below, I think you won't need this trim.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+                }
+                else
+                {
+                    result.Add(word);
+                }
+            }
+            return result;
+        }
+
+        private List<string> CleanWordsFromSymbols(List<string> words)
```
Luburic, line 427: What does it mean to clean words from symbols?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,404 @@ private int CountOccurrences(string text, string pattern)
...
+                        j++;
+                    }
+                    linesToRemove.Add(lines[i + j]);
+                }
+            }
+            lines.RemoveAll(line => linesToRemove.Contains(line));
+            return lines;
+        }
+
+        private List<string> GetWordsFromLines(List<string> lines)
```
Luburic, line 316: The code below needs work. We can discuss this tomorrow.

Luburic (changes requested):
Comments inline.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
simsimkic: That is how it was implemented in the https://github.com/mauricioaniche/ck, so I followed that example.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
simsimkic: The current implementation of the method does not include coalesce operators (?? and ?? =). CoalesceExpression would catch this example "b = a ?? -1;", and CoalesceAssignmentExpression would catch this example "b ?? = 2;". Should I include this?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
simsimkic: That is how it was implemented in the https://github.com/mauricioaniche/ck, so I followed that example.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
simsimkic: I could change the code so that it does not include string concatenation. The question is whether concatenation would actually complicate the method/class? If there is a "reasonable" number of concatenations, then probably not, but I could imagine a method with a large number of concatenations that would require the programmer to stop and take the time to interpret the meaning of the code.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
simsimkic: I was thinking of avoiding the block that represents the body of the method by extracting descendant nodes from the body (and in order to access the body, I had to cast). Now I see that it is not really the problem (the number of nested blocks is calculated correctly without it), so I will remove it.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
simsimkic: I wanted to count the words in the method body, so it was necessary to do the cast. As for the first check (whether a member is actually a method) - when all the tests are run, at some point a non-method parameter is passed, and the cast fails.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,394 @@ private int CountOccurrences(string text, string pattern)
...
+        private List<string> FilterWords(List<string> words)
+        {
+            for (int i = 0; i < words.Count(); i++)
+            {
+                words[i] = words[i]
+                    .Replace("(", " ")
+                    .Replace(")", " ")
+                    .Replace("{", " ")
+                    .Replace("}", " ")
+                    .Replace("=", " ")
```
Luburic, line 341: . i ?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +137,394 @@ private int CountOccurrences(string text, string pattern)
...
+                    {
+                        continue;
+                    }
+
+                    int current = 0;
+                    for (int i = 1; i < word.Length; i++)
+                    {
+                        if (word[i] == '_' || char.IsUpper(word[i]))
+                        {
+                            individualWords.Add(word.Substring(current, i - current).Trim());
```
Luburic, line 387: Split regex [A-Z], " ", "_"

RepositoryCompiler/CodeModel/CaDETModel/Metrics/CaDETClassMetrics.cs
```diff
@@ -9,5 +9,47 @@ public class CaDETClassMetrics
         public int WMC { get; set; }
         public int ATFD { get; set; }
         public double? TCC { get; set; }
+
+        /// <summary>
+        /// NOR: Number of return statements
+        /// Implementation based on https://github.com/mauricioaniche/ck
```
Luburic, line 15: These comments should be tied to the function that does the calculation - if we change the implementation, we won't be changing this class, but the metric calculator one.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
```diff
@@ -28,6 +28,7 @@ public string ContainerName
         public List<CaDETMember> Members { get; internal set; }
         public List<CaDETField> Fields { get; internal set; }
         public CaDETClassMetrics Metrics { get; internal set; }
+        public ISet<CaDETMember> InvokedMethods { get; internal set; }
```
Luburic, line 31: I wouldn't calculate this during parsing, nor store it in a field. There is no reason to maintain a set of class level invoked methods apart from the metric calculation - which is why we should move the code that is currently in the CodeParser to the MetricCalculator, get the set, count the elements, and let it be garbage collected.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -209,6 +209,11 @@ private List<CaDETClass> CalculateMetrics(List<CaDETClass> linkedClasses)
                     memberBuilder.CalculateMetrics();
                 }
                 c.Metrics = _metricCalculator.CalculateClassMetrics(c);
+                c.InvokedMethods = new HashSet<CaDETMember>();
+                foreach (var member in c.Members)
+                {
+                    c.InvokedMethods.UnionWith(member.InvokedMethods.ToList());
```
Luburic, line 215: Should be moved to the metric calculator. Even if we needed this set down the line, it would be better to have CaDETClass offer a method that dynamically creates this set.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETMemberMetricCalculator.cs
```diff
@@ -121,5 +141,296 @@ private int CountOccurrences(string text, string pattern)
...
+            words = FilterWords(words);
+            return words.Distinct().Count();
+        }
+
+        private List<string> GetWords(string methodBody)
+        {
+            List<string> words = new List<string>();
+            
+            int index = 0;
+            while (index < methodBody.Length && char.IsWhiteSpace(methodBody[index]))
```
Luburic, line 269: Can we not replace this code by calling split against " " (space), "\n" and "\r" and trimming each element?

Luburic (changes requested):
Comments inline.

Luburic (approved):
Looks good. Ready for merge.

simsimkic:

PR #17, opened 2021-03-25
Challenges

Ana00000:
First version of challenges.

### DIFF
--- a/SmartTutor/ContentModel/ChallengeService.cs
+++ b/SmartTutor/ContentModel/ChallengeService.cs
@@ -1,0 +1,20 @@
+﻿using SmartTutor.ContentModel.LearningObjects;
+using SmartTutor.ContentModel.LearningObjects.Repository;
+
+namespace SmartTutor.ContentModel
+{
+    public class ChallengeService : IChallengeService
+    {
+        private readonly ILearningObjectRepository _learningObjectRepository;
+
+        public ChallengeService(ILearningObjectRepository learningObjectRepository)
+        {
+            _learningObjectRepository = learningObjectRepository;
+        }
+
+        public Challenge GetChallenge(int challengeId)
+        {
+            return _learningObjectRepository.GetLearningObjectForChallenge(challengeId) as Challenge;
+        }
+    }
+}
--- a/SmartTutor/ContentModel/ChallengeService.cs
+++ b/SmartTutor/ContentModel/ChallengeService.cs
@@ -65,10 +65,10 @@ namespace SmartTutor.ContentModel
         {
             foreach (string metricName in submittetMetrics.Keys.ToList())
             {
-                Dictionary<string, double> resultMetrics = GetMetricsFromClasses(challenge.ResolvedClasses);
+                Dictionary<string, double> resultMetrics = GetMetricsFromClasses(challenge.ResolvedClasses);/*
                 if (resultMetrics[metricName] - challenge.MetricsRange[metricName] <= submittetMetrics[metricName]
-                    && submittetMetrics[metricName] <= resultMetrics[metricName] + challenge.MetricsRange[metricName])
-                    return true;
+                    && submittetMetrics[metricName] <= resultMetrics[metricName] + challenge.MetricsRange[metricName])*/
+                return true;
             }
 
             return false;
--- a/SmartTutor/ContentModel/ChallengeService.cs
+++ b/SmartTutor/ContentModel/ChallengeService.cs
@@ -19,15 +19,10 @@ namespace SmartTutor.ContentModel
         {
             List<CaDETClass> endState = GetChallenge(challengeId).EndState;
             List<CaDETClass> submittetState = GetClassesFromSubmittedChallenge(sourceCode);
-
-            // TODO: Implement rang for correct state, current is the simplest stage
-            if (endState != submittetState)
-                return false;
-
-            return true;
+            return CompareChallengeStates(endState, submittetState);
         }
 
-        private Challenge GetChallenge(int challengeId)
+        public Challenge GetChallenge(int challengeId)
         {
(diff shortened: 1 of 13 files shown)


Luburic (approved):
Looks good :)

SmartTutor/ContentModel/ChallengeService.cs
```diff
@@ -0,0 +1,105 @@
+﻿using RepositoryCompiler.CodeModel.CaDETModel.CodeItems;
+using RepositoryCompiler.Controllers;
+using SmartTutor.ContentModel.LearningObjects;
+using SmartTutor.ContentModel.LearningObjects.Repository;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SmartTutor.ContentModel
+{
+    public class ChallengeService : IChallengeService
```
Luburic, line 10: We should always keep in mind that the primary job of a service is to coordinate other (smarter) objects to collectively fulfill a use case.

SmartTutor/ContentModel/ChallengeService.cs
```diff
@@ -0,0 +1,105 @@
...
+        private readonly ILearningObjectRepository _learningObjectRepository;
+
+        public ChallengeService(ILearningObjectRepository learningObjectRepository)
+        {
+            _learningObjectRepository = learningObjectRepository;
+        }
+
+        public ChallengeService() { }
+
+        internal Dictionary<string, double> GetMetricNamesFromClasses(List<CaDETClass> caDETClasses)
```
Luburic, line 21: This type of logic is decidedly not coordination logic so we should think about what it does and where it fits better.

Notably, we will change the RepoCompiler CaDETModel so that the metrics and their values are key value pairs. Essentially, each CaDETClass will have a "Dictionary<string, double> Metrics".

Getting the names (althought I'm not sure why we need the names?) would then be trivial (e.g., Metrics.Keys)

SmartTutor/ContentModel/ChallengeService.cs
```diff
@@ -0,0 +1,105 @@
...
+                if (resultMetrics[metricName] - challenge.MetricsRange[metricName] <= submittetMetrics[metricName]
+                    && submittetMetrics[metricName] <= resultMetrics[metricName] + challenge.MetricsRange[metricName])
+                    return true;
+            }
+
+            return false;
+        }
+
+        // TODO: Developer might make more/less classes or members than there is in endState, change for those cases
+        private Dictionary<string, double> GetMetricsFromClasses(List<CaDETClass> caDETClasses)
```
Luburic, line 78: We should discuss this class.

Luburic (changes requested):
We should discuss the logic for validating challenges. There are also a few minor improvements I've commented on inline so check those out.

SmartTutor/ContentModel/LearningObjects/Repository/LearningObjectInMemoryRepository.cs
```diff
@@ -68,6 +126,9 @@ public List<LearningObject> GetFirstLearningObjectsForSummaries(List<int> summar
             return summaries.Select(id => _learningObjectCache[id].First()).ToList();
         }
 
-        
+        public LearningObject GetLearningObjectForChallenge(int challengeId)
+        {
+            return _learningObjectCache.SelectMany(LOs => LOs.Value.Where(LO => LO.Id == challengeId)).First();
```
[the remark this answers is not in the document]
Ana00000: That way I get the error for values of dictionary.

SmartTutor/ContentModel/LearningObjects/Challenge.cs
```diff
@@ -0,0 +1,47 @@
...
+        public string Url { get; internal set; }
+        public List<CaDETClass> ResolvedClasses { get; internal set; }
+        public List<MetricRangeRule> ClassMetricRules { get; internal set; }
+        public List<MetricRangeRule> MethodMetricRules { get; internal set; }
+
+        // TODO: This is basic strategy, implement strategy pattern interface 
+        public bool CheckSubmittedChallengeCompletion(List<CaDETClass> submittetClasses)
+        {
+            List<CaDETMember> submittedMethods = GetMethodsFromClasses(submittetClasses);
+            if (!CheckClassesCompletion(submittetClasses) || !CheckMethodsCompletion(submittedMethods))
```
Luburic, line 19: Lines 19-21 can be transformed into "return CheckClassesCompletion(submittetClasses) && CheckMethodsCompletion(submittedMethods)".

Always pay attention when you are returning booleans based on a branch - usually you can return some variation of the conditional for the branch.

SmartTutor/ContentModel/LearningObjects/Challenge.cs
```diff
@@ -0,0 +1,47 @@
...
+                return false;
+            return true;
+        }
+
+        private List<CaDETMember> GetMethodsFromClasses(List<CaDETClass> caDETClasses)
+        {
+            return caDETClasses.SelectMany(c => c.Members.Where(m => m.Type.Equals(CaDETMemberType.Method))).ToList();
+        }
+
+        private bool CheckClassesCompletion(List<CaDETClass> caDETClasses)
```
Luburic, line 29: I'm not a fan of the names CheckClassesCompletion and CheckMethodsCompletion.

Maybe "ValidateClassMetricRules"?

Luburic (commented):
Looks good, a few minor comments inline.

I'll look to transform the RepoCompiler later today to handle the metric dictionary. After that you should merge these changes and modify your challenge and we are done with this first version.

Also, make sure to tested your challenge with the VS Code plugin and see if everything works as expected.

Luburic (approved):
Looks good. You can resolve the conflict with main and merge.

PR #19, opened 2021-03-29
Challenge fulfillment strategy

Ana00000:
- Allowing future implementations of different challenge completion strategies.

### DIFF
--- a/SmartTutor/ContentModel/LearningObjects/ChallengeModel/Challenge.cs
+++ b/SmartTutor/ContentModel/LearningObjects/ChallengeModel/Challenge.cs
@@ -1,9 +1,9 @@
 ﻿using RepositoryCompiler.CodeModel.CaDETModel.CodeItems;
-using SmartTutor.ContentModel.LearningObjects.MetricRules;
+using SmartTutor.ContentModel.LearningObjects.ChallengeModel.MetricRules;
 using System.Collections.Generic;
 using System.Linq;
 
-namespace SmartTutor.ContentModel.LearningObjects
+namespace SmartTutor.ContentModel.LearningObjects.ChallengeModel
 {
     public class Challenge : LearningObject
     {
--- a/SmartTutor/ContentModel/LearningObjects/ChallengeModel/Challenge.cs
+++ b/SmartTutor/ContentModel/LearningObjects/ChallengeModel/Challenge.cs
@@ -13,9 +13,9 @@ namespace SmartTutor.ContentModel.LearningObjects.ChallengeModel
         public List<MetricRangeRule> MethodMetricRules { get; internal set; }
         public ChallengeFulfillmentStrategy FulfillmentStrategy { get; internal set; }
 
-        public bool CheckChallengeFulfillment(List<CaDETClass> submittetClasses, Challenge challenge)
+        public bool CheckChallengeFulfillment(List<CaDETClass> submittetClasses)
         {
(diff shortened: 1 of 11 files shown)


SmartTutor/ContentModel/LearningObjects/ChallengeModel/Challenge.cs
```diff
@@ -0,0 +1,21 @@
...
+using SmartTutor.ContentModel.LearningObjects.ChallengeModel.MetricRules;
+using System.Collections.Generic;
+
+namespace SmartTutor.ContentModel.LearningObjects.ChallengeModel
+{
+    public class Challenge : LearningObject
+    {
+        public string Url { get; internal set; }
+        public List<CaDETClass> ResolvedClasses { get; internal set; }
+        public List<MetricRangeRule> ClassMetricRules { get; internal set; }
```
Luburic, line 12: The MetricRangeRules are part of the basic challenge fulfillment strategy. For example, if we want to have an "AdequateNameChecker" as an implementation of the FulfillmentStrategy, we would host the key words we want to see in some name in this NameChecker, not in the Challenge.

This would also enable you to avoid passing the Challenge itself to the strategy, reducing the param. list and the feature envy code smell (since the strategy is extracting data from the Challenge currently).

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/ChallengeFulfillmentStrategy.cs
```diff
@@ -0,0 +1,10 @@
+﻿using RepositoryCompiler.CodeModel.CaDETModel.CodeItems;
+using System.Collections.Generic;
+
+namespace SmartTutor.ContentModel.LearningObjects.ChallengeModel.FulfillmentStrategy
+{
+    public abstract class ChallengeFulfillmentStrategy
+    {
+        public abstract bool CheckChallengeFulfillment(List<CaDETClass> submittetClasses, Challenge challenge);
```
Luburic, line 8: We can remove the challenge, and I would rename this list of classes (that is passed around in several places) to solutionAttempt or challengeSolutionAttempt.

submittedClasses is a bit redundant - first it is clear they are classes from the type, and submitted is half-informative since this data arrives on an endpoint through a POST request, making it submitted.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/BasicChallengeFulfillment.cs
```diff
@@ -1,45 +1,39 @@
...
-using SmartTutor.ContentModel.LearningObjects.MetricRules;
+using SmartTutor.ContentModel.LearningObjects.ChallengeModel.MetricRules;
 using System.Collections.Generic;
 using System.Linq;
 
-namespace SmartTutor.ContentModel.LearningObjects
+namespace SmartTutor.ContentModel.LearningObjects.ChallengeModel.FulfillmentStrategy
 {
-    public class Challenge : LearningObject
+    public class BasicChallengeFulfillment : ChallengeFulfillmentStrategy
```
Luburic, line 8: I would rename this class to "BasicMetricsChecker"

Luburic (changes requested):
Some renaming and moving a few fields, nothing big. Feel free to merge when you correct these changes, but make sure you've renamed all the appropriate fields.

PR #20, opened 2021-03-30
Metric hints

Ana00000:
- Creating metric hints for unfulfilled metric rules.


SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/BasicMetricsChecker.cs
```diff
@@ -29,20 +33,49 @@ private List<CaDETMember> GetMethodsFromClasses(List<CaDETClass> caDETClasses)
...
+            if (flag == 0)
+                return true;
+            return false;
+        }
+
+        private void AddHintForUnfulfilledMetricRule(MetricRangeRule metricRule)
+        {
+            _metricHints.Add(new MetricHint
+            {
+                Content = "Metric rule " + metricRule.MetricName + " should be between " + metricRule.FromValue + " and " + metricRule.ToValue + "."
```
Luburic, line 77: This isn't really a hint so much as the solution :)

I've been thinking about a more generalized ChallengeHint class that might even include a learning object. Basically, it would have the Content field like now + a reference to a LearningObject it might load (as a reminder). Maybe we should have the abstract strategy class contain a list of ChallengeHints and the implementation can decide the rules for when to return a specific hint. This will enable the trainee to ask for hints before they submit a challenge.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/BasicMetricsChecker.cs
```diff
@@ -29,20 +33,49 @@ private List<CaDETMember> GetMethodsFromClasses(List<CaDETClass> caDETClasses)
...
-                        return false;
-            return true;
+                    {
+                        flag++;
+                        AddHintForUnfulfilledMetricRule(classMetricRule);
+                    }
+                }
+            }
+            if (flag == 0)
+                return true;
```
Luburic, line 50: First, you can replace this code with "return flag == 0".

Second, I'm not sure we should be returning booleans anymore. Since we want to highlight specific hints, maybe we can return a ChallengeValidationResult object that contains the status of the challenge (e.g., Success, Fail) and a list of hints that are relevant for the solutionAttempt.

Luburic (commented):
A few comments inline. We can discuss this design further tomorrow.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/BasicMetricsChecker.cs
Ana00000: Yes, I agree. That is put as TODO and is next step for this PR.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/BasicMetricsChecker.cs
Ana00000: Understood! All minor questions I will ask tomorrow.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/MetricChecker/BasicMetricsChecker.cs
```diff
@@ -11,14 +12,18 @@ public class BasicMetricsChecker : ChallengeFulfillmentStrategy
...
+
+        public BasicMetricsChecker(List<MetricRangeRule> classMetricRules, List<MetricRangeRule> methodMetricRules, List<MetricHint> metricHints)
         {
             ClassMetricRules = classMetricRules;
             MethodMetricRules = methodMetricRules;
+            _metricHints = metricHints;
         }
 
+        //TODO: return object (with hints), not just bool
         public override bool CheckChallengeFulfillment(List<CaDETClass> solutionAttempt)
```
Luburic, line 27: I would return an object of type "ChallengeEvaluation" with the following fields:
- bool: ChallengeCompleted
- List<ChallengeHint> ApplicableHints
 (note the name - ChallengeHint)

When a challenge is complete, we can offer all hints if the trainee wants to examine them. Otherwise, we can offer hints that make sense. I've been thinking about who can offer these hints best, and it might be the low-level rules (in this case the MetricRule). For example, a MetricRule that counts the number of magic numbers might provide a hint for removing magic numbers. Basically, the MetricRule should return a Hint, and the BasicMetricChecker can aggregate these hints into a ChallengeEvaluation.
Does this make sense?

For the NameCheckerStrategy, we can define NamingRules that each contain a list of words to avoid and words to include. Each rule can be focused to return a Hint when it is not satisfied.

Luburic (changes requested):
Comments inline.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/MetricChecker/BasicMetricsChecker.cs
Ana00000: It does. I have separated the two types of hints: applicable and metric based. Hints that are being generated are moved to MetricRule, and I will upgrade them so they can give appropriate assistance for trainees.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/ChallengeHint.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SmartTutor.ContentModel.LearningObjects.ChallengeModel.FulfillmentStrategy
+{
+    public class ChallengeHint
+    {
+        public string Content { get; internal set; }
+        public int LearningObjectSummaryId { get; set; }
```
Luburic, line 6: This can be an LOSummary object.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/MetricHints/MetricHint.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace SmartTutor.ContentModel.LearningObjects.ChallengeModel.MetricHints
+{
+    public class MetricHint
```
Luburic, line 3: We don't need MetricHints as a separate class. The ChallengeHints are enough as they present a higher level of abstraction and hold the same data (and even more).

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/MetricChecker/MetricRangeRule.cs
```diff
@@ -12,10 +12,389 @@ public class MetricRangeRule
...
         public double ToValue { get; internal set; }
 
-        internal bool MetricMeetsRequirements(Dictionary<CaDETMetric, double> metrics)
+        public bool MetricMeetsRequirements(Dictionary<CaDETMetric, double> metrics)
         {
             var metricValue = metrics[(CaDETMetric)Enum.Parse(typeof(CaDETMetric), MetricName, true)];
             return (FromValue <= metricValue && metricValue <= ToValue);
         }
+
+        public ChallengeHint GetHintForIncompleteClass()
```
Luburic, line 21: It seems I didn't communicate well enough the intent of the ChallengeHint Content field. Basically, we would want this to be some additional information that is not present in the learning object, which the guy creating the challenge would manually populate in the database.

I feel bad saying this, but we really don't need the code below. First, it is hardcoded with English words. Second, most trainees won't know what the more elaborate metrics mean so they won't be sure how to comply with the advice. Third, advice such as "decrease the cohesion in the class" is not something we would ever recommend, while advice such as "increase the number of protected fields" is not a metric rule we would define. Fourth, every time we would add a metric we would need to modify these rules in the Smart Tutor even though our existing MetricRangeRules don't require the new metric.

The takeaway is - each MetricRangeRule has zero or one ChallengeHint associated with them. The strategy can then see if the rule passes, and if not see if the rule has a hint it can offer. This will be similar for the NamingRules.

If this requires further clarification we can talk over Teams.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/MetricChecker/BasicMetricsChecker.cs
```diff
@@ -33,19 +40,89 @@ private List<CaDETMember> GetMethodsFromClasses(List<CaDETClass> caDETClasses)
...
-                    if (!methodMetricRule.MetricMeetsRequirements(caDETMethod.Metrics))
-                        return false;
+                {
+                    if (!methodMetricRule.MetricMeetsRequirements(caDETMethod.Metrics)) return false;
+                }
+            }
             return true;
         }
+
+        private List<ChallengeHint> GetChallengeHints()
```
Luburic, line 64: The code below seems to have a lot of unneeded logic. Make sure you go over my comment below in the MetricRangeRuleClass first.

Then consider redesigning the logic in this class to have the following structure:
public override ChallengeEvaluation CheckChallengeFulfillment(List<CaDETClass> solutionAttempt)
{
    List<ChallengeHint> hints = new List<>();
    hints.AddRange(ValidateClassMetricRules(solutionAttempt));
    hints.AddRange(ValidateMethodMetricRules(solutionAttempt));
    if(hints.IsEmpty()) return new ChallengeEvaluation(true, GetAllHints());
    return new ChallengeEvaluation(false, hints);
}

The ValidateXMetricRules functions will create an empty hint list, go over the list of rules and for each rule that does not pass, ask it if it has a Hint - if yes add it, if no ignore it.

Luburic (changes requested):
A few major comments. Feel free to ping me if you have any questions.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/MetricHints/MetricHint.cs
Ana00000: This was accidentally left during last rebase.  This will be removed.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/MetricChecker/BasicMetricsChecker.cs
Ana00000: What happens when there is no hint but metric value is not appropriate? If there is no hint

> The code below seems to have a lot of unneeded logic. Make sure you go over my comment below in the MetricRangeRuleClass first.
> 
> Then consider redesigning the logic in this class to have the following structure:
> public override ChallengeEvaluation CheckChallengeFulfillment(List solutionAttempt)
> {
> List hints = new List<>();
> hints.AddRange(ValidateClassMetricRules(solutionAttempt));
> hints.AddRange(ValidateMethodMetricRules(solutionAttempt));
> if(hints.IsEmpty()) return new ChallengeEvaluation(true, GetAllHints());
What happens when there is no hint but metric value is not appropriate? Is tutor required to put hint for every metric rule or is there need for another validation which returns bool?

SmartTutor/ContentModel/ChallengeService.cs
```diff
@@ -27,6 +28,16 @@ public Challenge GetChallenge(int challengeId)
             return _learningObjectRepository.GetChallenge(challengeId);
         }
 
+        public List<ChallengeHint> GetAllHints(int challengeId)
+        {
+            return GetChallenge(challengeId).GetAllChallengeHints();
+        }
+
+        public List<ChallengeHint> GetHintsForSolutionAttempt(int challengeId, List<CaDETClass> solutionAttempt)
```
Luburic, line 36: As far as I can tell, this function supports the same Use case as the CheckChallengeCompletion function. We supply sourceCode and a challengeId and get an evaluation of our work. The only difference is that this does not include a bool for challenge completion. There is no need for this method.

SmartTutor/ContentModel/IChallengeService.cs
```diff
@@ -1,10 +1,15 @@
...
 
 namespace SmartTutor.ContentModel
 {
     public interface IChallengeService
     {
-        bool CheckChallengeCompletion(string[] sourceCode, int challengeId);
+        ChallengeEvaluation CheckChallengeCompletion(string[] sourceCode, int challengeId);
         Challenge GetChallenge(int challengeId);
+        List<ChallengeHint> GetAllHints(int challengeId);
+        List<ChallengeHint> GetHintsForSolutionAttempt(int challengeId, List<CaDETClass> solutionAttempt);
```
Luburic, line 13: Remove it from here as well.

SmartTutor/ContentModel/LearningObjects/ChallengeModel/FulfillmentStrategy/ChallengeFulfillmentStrategy.cs
```diff
@@ -1,12 +1,21 @@
...
 {
     public abstract class ChallengeFulfillmentStrategy
     {
-        [Key] public int Id { get; set; }
-        public abstract bool CheckChallengeFulfillment(List<CaDETClass> solutionAttempt);
+        [Key] public int Id { get; private set; }
+        public List<ChallengeHint> ChallengeHints { get; internal set; }
+
+        public abstract ChallengeEvaluation CheckChallengeFulfillment(List<CaDETClass> solutionAttempt);
+        public abstract List<ChallengeHint> GetHintsForSolutionAttempt(List<CaDETClass> submittedClasses);
```
Luburic, line 14: Remove. The children should decide how they will approach constructing the ChallengeEvaluation - no need to expose such a method as a public API.

SmartTutor/Controllers/DTOs/Challenge/BasicMetricsCheckerDTO.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace SmartTutor.Controllers.DTOs.Challenge
+{
+    public class BasicMetricsCheckerDTO
```
Luburic, line 5: Why do we need a DTO for the rules?

SmartTutor/Controllers/DTOs/Challenge/ChallengeFulfillmentStrategyDTO.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+
+namespace SmartTutor.Controllers.DTOs.Challenge
+{
+    public class ChallengeFulfillmentStrategyDTO
```
Luburic, line 5: DTOs are Data transfer objects - they are used to move data from the backend to a front-end application such as a the Web or Plugin. The challenge fulfillment logic and all its composing elements are backend logic that should never go to a front-end as it defeats the purpose of the challenge if you know the validation rules.

SmartTutor/Controllers/DTOs/Challenge/MetricRangeRuleDTO.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace SmartTutor.Controllers.DTOs.Challenge
+{
+    public class MetricRangeRuleDTO
```
Luburic, line 3: As mentioned, we don't need this.

SmartTutor/Controllers/DTOs/Lecture/ChallengeDTO.cs
```diff
@@ -1,8 +1,11 @@
-﻿namespace SmartTutor.Controllers.DTOs.Lecture
+﻿using SmartTutor.Controllers.DTOs.Challenge;
+
+namespace SmartTutor.Controllers.DTOs.Lecture
 {
     public class ChallengeDTO : LearningObjectDTO
     {
         public string Url { get; set; }
         public string Description { get; set; }
+        public ChallengeFulfillmentStrategyDTO FulfillmentStrategy { get; set; }
```
Luburic, line 9: Remove.

SmartTutor/SmartTutor.csproj
```diff
@@ -2,6 +2,9 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
+  <ItemGroup>
```
Luburic, line 5: Remove.

SmartTutorTests/Unit/ChallengeServiceTests.cs
```diff
@@ -1,31 +1,59 @@
...
-
-            Challenge challenge = challengeService.GetChallenge(3371);
+            Challenge challenge = _challengeService.GetChallenge(3371);
 
             challenge.Id.ShouldBe(3371);
             challenge.LearningObjectSummaryId.ShouldBe(337);
             challenge.Url.ShouldBe("https://github.com/Ana00000/Challenge-inspiration.git");
-            challenge.FulfillmentStrategy.ShouldNotBe(null);
+
+            BasicMetricsChecker basicMetricsChecker = (BasicMetricsChecker)challenge.FulfillmentStrategy;
```
Luburic, line 28: Remove.

Luburic (changes requested):
We should focus on cleanup now without adding additional content as the PR is mostly done. Address the inline comments and remove the recent additions so we can merge.

PR #21, opened 2021-03-30
CBO metric (number of dependencies)

simsimkic:
CBO metric counts the number of dependencies a class has. The calculation checks for any type used in the entire class:
- [x] field declaration
- [x] method return types
- [x] variable declarations

New model classes:
- CaDETVariable
- CaDETLinkedType

CaDETField is extended with:
- type

CaDETMember is extended with:
- return type
- list of variables


RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -105,7 +106,8 @@ private List<CaDETField> ParseFields(IEnumerable<MemberDeclarationSyntax> nodeMe
                     {
                         Name = field.Identifier.Text,
                         Parent = parent,
-                        Modifiers = GetModifiers(node)
+                        Modifiers = GetModifiers(node),
+                        Type = new CaDETClass() { Name = fieldDeclaration.Declaration.Type.ToString()}
```
simsimkic, line 110: I used the Name attribute (not FullName), because when declaring fields, type is just _name_ of the type (not _namespace.name_).

simsimkic:

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
```diff
@@ -26,6 +26,9 @@ public string ContainerName
         public List<CaDETMember> Members { get; internal set; }
         public List<CaDETField> Fields { get; internal set; }
         public Dictionary<CaDETMetric, double> Metrics { get; internal set; }
+        public List<CaDETClass> FieldTypes { get; internal set; }
```
Luburic, line 29: This can be calculated fields (i.e., methods) instead of having duplicated data. E.g., GetFieldTypes returns Members.Fields.SelectMany(f => f.Type);

This saves memory and processing power, since we will only sometimes want to access a class's fieldTypes and when we do we will call this only once.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETMember.cs
```diff
@@ -15,7 +15,9 @@ public class CaDETMember
         public ISet<CaDETMember> InvokedMethods { get; internal set; }
         public ISet<CaDETMember> AccessedAccessors { get; internal set; }
         public ISet<CaDETField> AccessedFields { get; internal set; }
+        public CaDETClass ReturnType { get; internal set; }
         public List<string> VariableNames { get; internal set; }
```
Luburic, line 19: VariableName and VariableType should be encapsulated into a CaDETVariable. Note that this will cause compilation errors in the BasicNamingStrategy of the SmartTutor so modify that code as well and run tests for both RC and ST.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCaDETMemberBuilder.cs
```diff
@@ -34,7 +34,9 @@ internal CaDETMember CreateBasicMember(CaDETClass parent)
             _member.SourceCode = _cSharpMember.ToString();
             _member.Parent = parent;
             _member.Params = GetMethodParams();
+            _member.ReturnType = GetMethodReturnType();
             _member.VariableNames = GetMethodVariableNames();
```
Luburic, line 38: GetMethodVariables to determine both.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
Luburic: So long as you are sure this will uniquely identify the type. Did you consider situations where there are multiple classes with the same name (but in a different namespace)?

RepositoryCompilerTests/Unit/MetricCalculationTests.cs
```diff
@@ -163,6 +163,22 @@ public void Calculates_max_nested_blocks_in_class()
             firstClass.Metrics[CaDETMetric.CMNB].ShouldBe(4);
         }
 
+        [Fact]
+        public void Calculates_number_of_dependencies_in_class()
```
Luburic, line 167: Since you added fields, variables, and return types, you should expand the other TestSuite in this package to check if your types are correctly determined. Make sure your tests include classes with the same name.

At one point we will need to refactor RepoCompiler tests to have better test data (the factory has gotten out of hand) - e.g., a complete project that we can thoroughly test with all edge cases.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -188,5 +189,16 @@ private int CountUniqueMethodInvocations(CaDETClass parsedClass)
...
+
+        // Implementation based on https://github.com/mauricioaniche/ck
+        private int CountDependencies(CaDETClass parsedClass)
+        {
+            List<CaDETClass> allDependencies = new List<CaDETClass>();
+            allDependencies.AddRange(parsedClass.FieldTypes);
+            allDependencies.AddRange(parsedClass.MethodReturnTypes);
+            allDependencies.AddRange(parsedClass.MethodVariableTypes);
+            var uniqueDependencies = allDependencies.GroupBy(d => d.FullName).Select(d => d.First());
+            return uniqueDependencies.ToList().Count();
```
Luburic, line 201: Enumerable has a Count method.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -199,6 +204,78 @@ private CaDETClass LinkOuterClass(List<CaDETClass> classes, string containerName
         {
             return classes.FirstOrDefault(c => c.FullName.Equals(containerName));
         }
+        
+        private List<CaDETClass> LinkFieldTypes(List<CaDETClass> classes, List<CaDETField> fields)
+        {
+            List<CaDETClass> fieldTypes = new List<CaDETClass>();
+            foreach (var f in fields)
```
Luburic, line 211: Can this loop be replaced with fieldTypes.AddRange(FindTypes(classes, fields.Select(f => f.Type)))?

In light of the above comment I see that this approach is not suitable. We need to change each field to set a new Type for it.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -199,6 +204,78 @@ private CaDETClass LinkOuterClass(List<CaDETClass> classes, string containerName
...
+            return returnTypes.Where(t => t != null).Distinct().ToList();
+        }
+
+        private List<CaDETClass> LinkMethodVariableTypes(List<CaDETClass> classes, List<CaDETMember> members)
+        {
+            var methods = members.Where(m => m.Type == CaDETMemberType.Method).ToList();
+            List<CaDETClass> variableTypes = new List<CaDETClass>();
+            foreach (var m in methods)
+            {
+                variableTypes.AddRange(FindTypes(classes, m.VariableTypes));
```
Luburic, line 234: SelectMany

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -199,6 +204,78 @@ private CaDETClass LinkOuterClass(List<CaDETClass> classes, string containerName
...
+            foreach (var m in methods)
+            {
+                returnTypes.AddRange(FindTypes(classes, new List<CaDETClass>() { m.ReturnType}));
+            }
+            return returnTypes.Where(t => t != null).Distinct().ToList();
+        }
+
+        private List<CaDETClass> LinkMethodVariableTypes(List<CaDETClass> classes, List<CaDETMember> members)
+        {
+            var methods = members.Where(m => m.Type == CaDETMemberType.Method).ToList();
```
Luburic, line 230: We are ignoring constructors?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -182,6 +184,9 @@ private List<CaDETClass> ConnectCaDETGraph(List<CaDETClass> classes)
             {
                 c.Parent = LinkParent(classes, c.Parent);
                 c.OuterClass = LinkOuterClass(classes, c.ContainerName);
+                c.FieldTypes = LinkFieldTypes(classes, c.Fields);
```
Luburic, line 187: You aren't really linking the types. You are generating the collections for the class, but the fields, variables, and return type of a method are not changed, meaning that our model is populated with dummy references that can't be used.

Once you remove these fields from the class the problem will present itself.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -199,6 +204,78 @@ private CaDETClass LinkOuterClass(List<CaDETClass> classes, string containerName
...
+            }
+            return fieldTypes.Where(t => t != null).Distinct().ToList();
+        }
+        private List<CaDETClass> LinkMethodReturnTypes(List<CaDETClass> classes, List<CaDETMember> members)
+        {
+            var methods = members.Where(m => m.Type == CaDETMemberType.Method).ToList();
+            List<CaDETClass> returnTypes = new List<CaDETClass>();
+            foreach (var m in methods)
+            {
+                returnTypes.AddRange(FindTypes(classes, new List<CaDETClass>() { m.ReturnType}));
```
Luburic, line 223: Same as above.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -199,6 +204,78 @@ private CaDETClass LinkOuterClass(List<CaDETClass> classes, string containerName
...
+            }
+            return variableTypes.Where(t => t != null).Distinct().ToList();
+        }
+
+        private List<CaDETClass> FindTypes(List<CaDETClass> classes, List<CaDETClass> types)
+        {
+            List<CaDETClass> foundTypes = new List<CaDETClass>();
+            foreach (var type in types)
+            {
+                foundTypes.Add(classes.FirstOrDefault(c => c.Name.Equals(type.Name)));
```
Luburic, line 244: This foreach will add a bunch of nulls, which is why you need to repeat the Where(t => t != null) in all of the above methods instead of doing it once at the end of this method.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -199,6 +204,78 @@ private CaDETClass LinkOuterClass(List<CaDETClass> classes, string containerName
...
+        }
+
+        private List<CaDETClass> GetCollectionTypes(List<CaDETClass> classes, string type)
+        {
+            List<CaDETClass> collectionTypes = new List<CaDETClass>();
+            var match = Regex.Match(type, "<.+>");
+            if (!match.Success) return collectionTypes;
+
+            var value = match.Value.Substring(1, match.Value.Length - 2);
+            var types = value.Split(",").Select(t => t.Trim()).ToList();
```
Luburic, line 258: Longer variable names would be beneficial to more clearly express the intent of each segment of the code.

"value" could be "typeParameterSection" using the terminology from https://docs.oracle.com/javase/tutorial/java/generics/types.html

"types" could be typeParameters

Luburic (changes requested):
Comments inline.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETClass.cs
simsimkic: Member nema Fields?

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
simsimkic: Changed - full name is taken into account.

RepositoryCompiler/CodeModel/CaDETModel/CodeItems/CaDETLinkedType.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+
+namespace RepositoryCompiler.CodeModel.CaDETModel.CodeItems
+{
+    public class CaDETLinkedType
```
Luburic, line 5: A summary comment would be helpful here. It should contain a brief description of what the linkedType is and an example or two.

RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -88,24 +89,26 @@ private CaDETClass ParseClass(SemanticModel semanticModel, ClassDeclarationSynta
...
         {
             List<CaDETField> fields = new List<CaDETField>();
+
             foreach (var node in nodeMembers)
             {
                 if (!(node is FieldDeclarationSyntax fieldDeclaration)) continue;
+                
                 fields.AddRange(fieldDeclaration.Declaration.Variables.Select(
-                    field => new CaDETField
+                    variable => new CaDETField
```
Luburic, line 106: Why is this renamed to "variable"?

RepositoryCompilerTests/Unit/CaDETCodeModelTests.cs
```diff
@@ -212,16 +212,80 @@ public void Calculates_method_variables()
...
+            List<CaDETClass> classes = factory.CreateClassModel(_testDataFactory.GetMultipleClassTexts());
+            List<CaDETClass> otherClasses = factory.CreateClassModel(_testDataFactory.GetClassesFromDifferentNamespace());
+
+            var dateRange = classes.Find(c => c.Name.Equals("DateRange"));
+            var doctor = classes.Find(c => c.Name.Equals("Doctor"));
+            var doctorService = classes.Find(c => c.Name.Equals("DoctorService"));
+            var otherDoctor = otherClasses.Find(c => c.Name.Equals("Doctor"));
+            var otherDateRange = otherClasses.Find(c => c.Name.Equals("DateRange"));
+
+            doctor.FindMember("TestFunction").Variables.SelectMany(v => v.GetLinkedTypes()).ShouldContain(doctor);
```
Luburic, line 284: doctor.FindMember("TestFunction").Variables.SelectMany(v => v.GetLinkedTypes()) is duplicated 5 times. Change to:

var variables = doctor.FindMember("TestFunction").Variables.SelectMany(v => v.GetLinkedTypes());
variables.ShouldContain...

Luburic (commented):
Several minor comments inline to address.

Make sure you run RepoCompiler unit and integration tests (for the projects we processed so far), as well as the SD, ST, and DE tests which shouldn't be affected, but still to be safe.

Luburic (approved):
Looks good.

PR #22, opened 2021-04-02
Name checker strategy

Ana00000:
- Basic strategy for required and banned words in challenges.


Luburic (changes requested):
Comments inline.

PR #23, opened 2021-04-03
Recommender flow and knowledge based recommender

milicasiriski:


SmartTutor/ContentModel/ContentService.cs
```diff
@@ -63,13 +66,16 @@ public NodeProgress GetNodeContent(int knowledgeNodeId, int? traineeId)
 
         private NodeProgress CreateNodeForTrainee(int knowledgeNodeId, int? traineeId)
```
Luburic, line 67: We can change the signature to have a mandatory traineeId, and add an explicit (int) cast in line 51.

SmartTutor/ContentModel/ContentService.cs
```diff
@@ -63,13 +66,16 @@ public NodeProgress GetNodeContent(int knowledgeNodeId, int? traineeId)
...
         {
-            //TODO: Load Trainee prefs
-            //TODO: Get recommender to build NodeProgress with LOs for Trainee
-            //TODO: Save started NodeProgress to repo
+            if (traineeId == null) return null; // TODO: throw an exception?
+            var trainee = _traineeRepository.GetTraineeById(traineeId.Value);
+            var knowledgeNode = _lectureRepository.GetKnowledgeNodeWithSummaries(knowledgeNodeId);
+
+            var nodeProgress = _recommender.BuildNodeProgressForTrainee(trainee, knowledgeNode);
+            _traineeRepository.SaveNodeProgress(nodeProgress);
```
Luburic, line 74: This scenario assumes the Trainee did not already start the Node. We should also consider when the Trainee has already started/finished a node - such a node would be loaded instead of constructed. We can address this scenario next, not in this PR.

SmartTutorTests/Unit/NodeContentTests.cs
```diff
@@ -12,9 +12,10 @@ namespace SmartTutorTests.Unit
     public class NodeContentTests
     {
         private readonly ContentService _service;
+
         public NodeContentTests()
         {
-            _service = new ContentService(null, null, CreateMockRepository());
+            _service = new ContentService(null, null, CreateMockRepository(), null);
```
Luburic, line 18: Discard the formatting changes in this file and only add the null.

SmartTutor/Recommenders/KnowledgeBasedRecommender.cs
```diff
@@ -20,5 +27,39 @@ public List<LearningObject> FindEducationalContent(List<SmellType> issues)
...
+                Trainee = trainee, Node = knowledgeNode, Status = NodeStatus.Started, LearningObjects = learningObjects
+            };
+        }
+
+        private LearningObject GetLearningObjectForPreference(LearningPreference learningPreference, int summaryId)
+        {
+            return learningPreference switch
+            {
+                LearningPreference.Aural => _learningObjectRepository.GetVideoForSummary(summaryId),
+                LearningPreference.Kinaesthetic => _learningObjectRepository.GetQuestionForSummary(summaryId),
```
Luburic, line 58: ArrangeTask is also an interactive LO. I would change the repo function to GetInteractiveLOForSummary.

SmartTutor/Recommenders/KnowledgeBasedRecommender.cs
```diff
@@ -20,5 +27,39 @@ public List<LearningObject> FindEducationalContent(List<SmellType> issues)
...
+        public NodeProgress BuildNodeProgressForTrainee(Trainee trainee, KnowledgeNode knowledgeNode)
+        {
+            var learningObjects = new List<LearningObject>();
+            var sortedPreferences = trainee.LearningPreferenceScore().OrderByDescending(key => key.Value).ToList();
+
+            foreach (var summary in knowledgeNode.LearningObjectSummaries)
+            {
+                foreach (var learningObject in sortedPreferences
+                    .Select(preference => GetLearningObjectForPreference(preference.Key, summary.Id))
+                    .Where(learningObject => learningObject != null))
```
Luburic, line 40: We should return one LO for each summary for now. As far as I can tell this code also won't return anything if there is only a challenge tied to a summary (or new types we will add in the future). We should always have a fallback - if you don't find anything, select the first available.

Luburic (changes requested):
Several comments inline to address. Also, we could use some tests, so build a Theory in a new RecommenderTests class that will have e.g. 3 trainees for a single knowledge node with a one or two summaries. You can utilize the InMemRepo for the Mock or build your own.

SmartTutor/ContentModel/LearningObjects/Repository/ILearningObjectRepository.cs
```diff
@@ -10,5 +10,11 @@ public interface ILearningObjectRepository
         Challenge GetChallenge(int challengeId);
         List<QuestionAnswer> GetQuestionAnswers(int questionId);
         List<ArrangeTaskContainer> GetArrangeTaskContainers(int arrangeTaskId);
+        Image GetImageForSummary(int summaryId);
+        LearningObject GetInteractiveLOForSummary(int summaryId);
+        Text GetTextForSummary(int summaryId);
+        Video GetVideoForSummary(int summaryId);
+        Challenge GetChallengeForSummary(int summaryId);
```
Luburic, line 17: There already is a GetChallenge function above, no need to add it for a summary specifically.

Luburic (commented):
Added one minor comment, while the other from last time still stands (regarding the node progress loading). We can discuss this today.

Luburic (approved):
Looks good.

PR #24, opened 2021-04-04
VariableNames property in CaDETMember

simsimkic:
- CaDETMember class expanded with VariableNames property (list of variable identifiers)


RepositoryCompiler/CodeModel/CodeParsers/CSharp/CSharpCaDETMemberBuilder.cs
```diff
@@ -112,6 +113,14 @@ private List<CaDETParameter> GetMethodParams()
             return memberParams;
         }
 
+        private List<string> GetMethodVariableNames()
+        {
+            var variables = _cSharpMember.DescendantNodes().OfType<VariableDeclaratorSyntax>().ToList();
+            List<string> variableNames = new List<string>();
```
Luburic, line 119: Lines 119-121 can be reduced to "return variables.Select(v => v.Identifier.ToString())"

Luburic (approved):
One minor comment to address and you can merge.

PR #28, opened 2021-04-08
Content service returns existing node progress (otherwise recommended learning objects are generated

milicasiriski:


SmartTutorTests/Unit/ContentServiceTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public static IEnumerable<object[]> TraineeTestData =>
+            new List<object[]>
+            {
+                new object[]
+                {
+                    1,
+                    1,
+                    0
```
Luburic, line 55: This test is too coupled to the implementation and fails to satisfy the property of a good test "resistant to refactoring".

We should be verifying if we are getting a new node or an existing node based on the trainee/node id combination. How that's implemented is of little concern.

Basically, the MockRepo should return an existing node progress (e.g., with ID 1) while the creation function should make a new one (e.g., ID 7).

SmartTutorTests/Unit/ContentServiceTests.cs
```diff
@@ -0,0 +1,69 @@
...
+                },
+                new object[]
+                {
+                    2,
+                    1,
+                    1
+                }
+            };
+
+        private static readonly Trainee Trainee = new Trainee {Id = 1};
```
Luburic, line 65: We should avoid populating this class with fields like this as it can quickly expand and become unmanageable. Inline the object instantiation into the appropriate methods.

Luburic (approved):
Minor comments related to tests that you can address and merge.

SmartTutorTests/Unit/ContentServiceTests.cs
milicasiriski: I ended up having to use these because of the issues with different references. I would have also had the same instantiations in different places.

PR #30, opened 2021-04-12
File tester

Ana00000:
- Checking if file, generated from source code, is executable.


Clean CaDET Platform.sln
```diff
@@ -17,7 +17,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SmartTutorTests", "SmartTut
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DataSetExplorer", "DataSetExplorer\DataSetExplorer.csproj", "{5064A050-8C9B-4AAC-8F5C-09BEDCFCF4C8}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DataSetExplorerTests", "DataSetExplorerTests\DataSetExplorerTests.csproj", "{E6AA4545-2AB6-42E0-B560-612F64667366}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DataSetExplorerTests", "DataSetExplorerTests\DataSetExplorerTests.csproj", "{E6AA4545-2AB6-42E0-B560-612F64667366}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SourceCodeTests", "SourceCodeTests\SourceCodeTests.csproj", "{3463662F-E9C5-4F5C-9981-A84775C10D14}"
```
Luburic, line 22: What is the reasoning behind a completely new project? Since this functionality is tied to the challenges, can we not include it in the SmartTutorTests?

Luburic (commented):
For now one question.

Apart from ensuring that a file is executable, can we run a set of unit tests on it?
Say I am refactoring function X. Refactoring by its definition is changing the internal structure without changing the visible behavior. A test that passed for function X before the refactoring should pass for function X after the refactoring. This would be a first check before applying the challenge evaluation strategies.

Clean CaDET Platform.sln
Ana00000: This was a misunderstanding in communication. This will be moved.

SmartTutor/ContentModel/ChallengeService.cs
```diff
@@ -18,6 +20,9 @@ public ChallengeService(ILearningObjectRepository learningObjectRepository)
 
         public ChallengeEvaluation EvaluateSubmission(string[] sourceCode, int challengeId)
         {
+            if (!new SourceCodeTester().ValidSourceCode(sourceCode))
+                throw new Exception("Code can not be executed!");
```
Luburic, line 24: InvalidStateException might be more appropriate.

SmartTutor/SmartTutor.csproj
```diff
@@ -18,5 +18,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\RepositoryCompiler\RepositoryCompiler.csproj" />
+    <ProjectReference Include="..\SourceCodeTests\SourceCodeTests.csproj" />
```
Luburic, line 21: Once you move the logic and delete the project this should be discarded as well.

As a side note - the main project should never reference the test project.

SourceCodeTests/ExecutableFileTests.cs
```diff
@@ -0,0 +1,38 @@
...
+            FileContent = File.ReadAllBytes(filePath);
+        }
+
+        [Fact]
+        public void Check_If_Executable_File()
+        {
+            byte[] result = new byte[2];
+            Array.Copy(FileContent, 0, result, 0, 2);
+
+            try
```
Luburic, line 25: Test logic should have 0 nesting - this means no branches, loops, or try-catch.

Shouldly supports testing if a function throws an exception - https://docs.shouldly.io/documentation/exceptions/throw#79612a9b

SourceCodeTests/FileStorage/SourceCodeFile.txt
```diff
@@ -0,0 +1,83 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace ExamplesApp.Method.Violation
```
Luburic, line 6: Can be .cs instead of .txt.

Luburic (changes requested):
Comments inline. In general, we should remove the new project and move the testing logic to the SmartTutorTests.

The SourceCodeTester can be a class in the ChallengeModel. The first check is to see if this is valid source code, the second check is to determine if it is functionally correct. We can talk more about the second check tomorrow.

PR #31, opened 2021-04-13
Trainee registration and login

Luburic:


SmartTutor/Controllers/TraineeController.cs
```diff
@@ -0,0 +1,44 @@
...
+        public ActionResult<TraineeDTO> LoginTrainee([FromBody] LoginDTO login)
+        {
+            try
+            {
+                var loggedInTrainee = _traineeService.LoginTrainee(login.StudentIndex);
+                return Ok(_mapper.Map<TraineeDTO>(loggedInTrainee));
+            }
+            catch (TraineeWIthStudentIndexNotFound)
+            {
+                return BadRequest();
```
Luburic, line 40: Maybe return NotFound(e.message)?

SmartTutor/Exceptions/TraineeWithStudentIndexAlreadyExists.cs
```diff
@@ -0,0 +1,8 @@
+using System;
+
+namespace SmartTutor.Exceptions
```
Luburic, line 3: Probably don't need this with the current design.

Luburic (commented):
A few minor comments.

PR #35, opened 2021-04-19
Annotated dataset file

simsimkic:
Generates a dataset file for each code smell type. Files contain instances of all annotated projects. The file contains:

1. Code snippet ID (full name of class/method)
2. Link (code snippet Github url)
3. Code smell (type of code smell)
4. Project link (project Github url)
5. Metrics (value for each calculated metric)
6. Annotations (value from each annotator)
7. Final annotation (based on individual annotations from annotators)

The final annotation is determined as follows:

- all annotators annotated the instance: the final annotation is the majority annotation
- other: the final annotation is the annotation of the most experienced annotator


simsimkic:

DataSetExplorer/DataSetBuilder/Model/ExportedDataSetInstance.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        public bool IsAnnotatedByMostExperienced()
+        {
+            return IsAnnotatedBy(MostExperiencedAnnotatorId);
+        }
+
+        public bool IsAnnotatedBy(int annotatorId)
+        {
+            return Annotations.Exists(a => a.AnnotatorId == annotatorId);
```
Luburic, line 58: I'm not sure but I think most of these methods are not needed. Find my comment below with Annotators 1 - 4 and consider moving the fields and logic to the DataSetInstance and/or DataSetAnnotation classes while deleting this class.

DataSetExplorer/DataSetSerializer/DataSetExporter.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        private void PopulateAnnotations(ExportedDataSetInstance instance, int row)
+        {
+            var numOfMetrics = instance.Metrics.Count;
+            var maxNumberOfAnnotations = ExportedDataSetInstance.MaxNumberOfAnnotations;
+
+            for (var i = 0; i < maxNumberOfAnnotations; i++)
+            {
+                var annotatorId = i + 1;
+                sheet.Cells[1, 5 + numOfMetrics + i].Value = "Annotator " + annotatorId;
```
Luburic, line 63: Ok for now, but it would be better if we had a more structured template so that the first row would define the column names (e.g., "Metrics" and "Annotator Ids") and the second row would then have the metric names and the actual annotator IDs.

DataSetExplorer/DataSetSerializer/DataSetExporter.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        private void PopulateAnnotations(ExportedDataSetInstance instance, int row)
+        {
+            var numOfMetrics = instance.Metrics.Count;
+            var maxNumberOfAnnotations = ExportedDataSetInstance.MaxNumberOfAnnotations;
+
+            for (var i = 0; i < maxNumberOfAnnotations; i++)
+            {
+                var annotatorId = i + 1;
```
Luburic, line 62: This only works if the annotator IDs are 1, 2, 3.

DataSetExplorer/Program.cs
```diff
@@ -27,8 +33,91 @@ class Program
         static void Main(string[] args)
```
Luburic, line 33: This class currently owns many use cases that should be exported to a few services that can perform the coordination for each use case. Not something to solve as part of this PR, but definitely one of the next steps.

DataSetExplorer/Program.cs
```diff
@@ -27,8 +33,91 @@ class Program
...
+            projects.Add("D:/ccadet/annotations/repos/OpenRA", "D:/ccadet/annotations/annotated/OpenRA");
+            projects.Add("D:/ccadet/annotations/repos/ShareX", "D:/ccadet/annotations/annotated/ShareX");
+            projects.Add("D:/ccadet/annotations/repos/ShopifySharp", "D:/ccadet/annotations/annotated/ShopifySharp");
+
+            Dictionary<string, List<ExportedDataSetInstance>> dataForExport = new Dictionary<string, List<ExportedDataSetInstance>>();
+
+            foreach (var key in projects.Keys)
+            {
+                CodeModelFactory factory = new CodeModelFactory(LanguageEnum.CSharp);
+                CaDETProject project = factory.CreateProjectWithCodeFileLinks(key.ToString());
```
Luburic, line 56: Feel free to add a "GetMetricsForCodeSnippet" method in the CaDETProject.

DataSetExplorer/Program.cs
```diff
@@ -27,8 +33,91 @@ class Program
...
+                {
+                    Metrics = GetMetrics(instance, project),
+                    Annotations = instance.Annotations.ToList()
+                };
+                dataForExport.Add(exportedDataSetInstance);
+            }
+            return dataForExport;
+        }
+
+        private static Dictionary<CaDETMetric, double> GetMetrics(DataSetInstance instance, CaDETProject project)
```
Luburic, line 106: Feature envy for CaDETProject.

DataSetExplorer/Program.cs
```diff
@@ -27,8 +33,91 @@ class Program
         static void Main(string[] args)
         {
             //MakeExcelFromProjectUseCase();
-            FindInstancesRequiringAdditionalAnnotationUseCase();
+            //FindInstancesRequiringAdditionalAnnotationUseCase();
             //FindInstancesWithAllDisagreeingAnnotationsUseCase();
+            ExportAnnotatedDataSet();
+        }
+        
+        private static void ExportAnnotatedDataSet()
```
Luburic, line 41: This method would benefit from a communication diagram to understand who is the coordinator and who is responsible for what.

DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
```diff
@@ -55,11 +55,36 @@ internal bool HasNoAgreeingAnnotations()
...
+            var majorityVote = GetMajorityAnnotation();
+            if (majorityVote != null) return (int)majorityVote;
+            return GetAnnotationFromMostExperiencedAnnotator();
+        }
+
+        private int? GetMajorityAnnotation()
+        {
+            var annotationsGroupedBySeverity = Annotations.GroupBy(a => a.Severity);
+            var severityCounts = annotationsGroupedBySeverity.Select(group => group.Count());
+            if (severityCounts.Any(count => count != severityCounts.First()))
```
Luburic, line 74: Intent of this statement is not clear - What is it checking? If there is more than one element in severityCounts that is not equal to first? How do we know the first element is the largest number?

DataSetExplorer/DataSetSerializer/AnnotatorData/annotator_data.csv
```diff
@@ -0,0 +1,4 @@
+﻿id;yearsOfExperience;ranking
+1;6;1
```
Luburic, line 2: This type of data shouldn't be in the source code unless it's test data (and while we're at it, the tests should also have this data somewhere, along with the BK test data).

For our purposes you can hard code it in Program.cs until we refactor it but don't introduce this file to the repo.

DataSetExplorer/DataSetSerializer/DataSetWithMetricsExporter.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        public DataSetWithMetricsExporter(string exportPath)
+        {
+            _exportPath = exportPath;
+            ExcelPackage.LicenseContext = LicenseContext.NonCommercial;
+            template = new ExcelPackage(new FileInfo(_templatePath));
+            sheet = template.Workbook.Worksheets[0];
+        }
+
+        public void Export(List<Tuple<DataSetInstance, Dictionary<CaDETMetric, double>>> dataSetInstances,
```
Luburic, line 27: If I'm not mistaken this method can only be called once with the current design. For example, say we persist the dataset that has 1000 rows, which will populate the sheet and save it to file A. Next we call another dataset that has 500 rows that saves to file B. The B also has a 1000 rows since the second 500 are still stored.

This means that the client code must always instantiate the Exporter and then call Export and be aware of this detail, breaking encapsulation. In general, we expect our objects to be able to handle multiple calls of their public API, so you should recreate the sheet for export purposes.

DataSetExplorer/DataSetSerializer/DataSetWithMetricsExporter.cs
```diff
@@ -0,0 +1,107 @@
...
+using OfficeOpenXml.Style;
+using RepositoryCompiler.CodeModel.CaDETModel.CodeItems;
+
+namespace DataSetExplorer.DataSetSerializer
+{
+    class DataSetWithMetricsExporter
+    {
+        private readonly string _templatePath = "../../../DataSetSerializer/Template/Dataset_Metrics_Template.xlsx";
+        private readonly string _exportPath;
+        private ExcelPackage template;
```
Luburic, line 16: private members start with _. Also "template" is a misleading name.

Luburic (changes requested):
Comments inline.

DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
simsimkic: It does not have to be the largest, it is just important to check how many votes each severity got (e.g. two votes for severity 2 and two votes for severity 3). The goal of this condition is to check if all severities are equally voted - if they are not, then I take the severity that has the most votes.

DataSetExplorer/DataSetSerializer/DataSetWithMetricsExporter.cs
simsimkic: Is the _excelFile ok?

DataSetExplorer/DataSetSerializer/DataSetWithMetricsExporter.cs
simsimkic: This method is called multiple times within Program.cs - it is currently called twice since we have two code smells. Each generated file has only as many rows as needed, there is no duplication or rows from previous files. The exporter is also instantiated as many times as needed (in Program.cs in ExportAnnotatedDataSet() method).

DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
Luburic: Got it, but it took me a while. I would extract the logic into a method, e.g.:

    if(HasMajoritySeverityVote(annotationsGroupedBySeverity)) { //includes the severityCounts creation.
        return annotationsGroupedBySeverity.Max(a => a.Count()).Key; //not sure if Max exists, so check.
    }

DataSetExplorer/Program.cs
```diff
@@ -27,8 +34,67 @@ class Program
...
+            projects.Add("D:/ccadet/annotations/repos/ShareX", "D:/ccadet/annotations/annotated/ShareX");
+            projects.Add("D:/ccadet/annotations/repos/ShopifySharp", "D:/ccadet/annotations/annotated/ShopifySharp");
+
+            var dataForExport = PrepareDataForExport(projects);
+            var groupedBySmells = dataForExport.GroupBy(t => t.Item1.Annotations.ToList()[0].InstanceSmell.Value);
+            
+            var enumerator = groupedBySmells.GetEnumerator();
+            while (enumerator.MoveNext())
+            {
+                var exporter = new DataSetWithMetricsExporter("D:/ccadet/annotations/annotated/Output/");
```
Luburic, line 58: Regarding your comment: Program.cs in this case is the client code of DataSetWithMetricsExporter. You are forcing the Program.cs to instantiate DataSetWithMetricsExporter twice, or rather once for each Export operation.

DataSetWithMetricsExporter would be more useful if it could be instantiated above the loop and then Export could be called multiple times.

Luburic (commented):
Additional info inline. Regarding _excelFile comment, it's an acceptable name.

DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
simsimkic: Max gives the max number of votes for a severity, so I can't get the key (the actual severity).

DataSetExplorer/DataSetBuilder/Model/DataSetAnnotation.cs
```diff
@@ -8,36 +8,36 @@ public class DataSetAnnotation
...
         }
 
-        public override int GetHashCode() => (AnnotatorId, InstanceSmell: InstanceSmell.Value).GetHashCode();
+        public override int GetHashCode() => (Annotator.Id, InstanceSmell: InstanceSmell.Value).GetHashCode();
 
         public override bool Equals(object other)
         {
             return other is DataSetAnnotation annotation
-                   && AnnotatorId.Equals(annotation.AnnotatorId)
+                   && Annotator.Id.Equals(annotation.Annotator.Id)
```
Luburic, line 34: This can be moved to the annotator class, so that we have Annotator.Equals(Annotator) here. Don't forget the hash code (this might help https://docs.microsoft.com/en-us/visualstudio/ide/reference/generate-equals-gethashcode-methods?view=vs-2019 or find a similar article).

DataSetExplorerTests/Unit/ImporterTests.cs
```diff
@@ -13,18 +16,50 @@ public class ImporterTests
...
+
+        private void JoinInstancesAndAnnotators(ref List<DataSetInstance> annotatedInstances, List<Annotator> annotators)
+        {
+            foreach (var annotation in annotatedInstances.SelectMany(i => i.Annotations))
+            {
+                annotation.Annotator = annotators.Find(annotator => annotator.Id.Equals(annotation.Annotator.Id));
+            }
+        }
+
+        private List<Annotator> ImportAnnotators()
```
Luburic, line 48: Should this logic be in the test? Can we simply mock this data and put it in the _arrange_ section of the test instead of coupling it to this file?

Luburic (commented):
A few minor comments to address and then you can merge.

Luburic (approved):

PR #36, opened 2021-04-21
Keycloak authorization and authentication setup.

lukaDoric:
Definition of done:
User is able to identify himself in SmartTutor with JWT that is gather from Keycloak.
Also we are able to create specific policy just for users from Keycloak, and if necessary it can be turned off.
image

I made an example in the picture above where we can see that only the user who has Administrator Role satisfies policy.

Audience:  If the value specified in this parameter doesn’t match the aud parameter in the token, the token will be rejected because it was meant to be used for accessing a different resource.
Authority: Is the address of the token-issuing authentication server. The JWT bearer authentication middleware will use this URI to find and retrieve the public key that can be used to validate the token’s signature.

These two parameters are empty for now but they can be configured like this (picture).
image

Useful links:
https://jakeydocs.readthedocs.io/en/latest/security/authorization/policies.html

### DIFF
--- a/SmartTutor/KeycloakAuth/AllowedKeycloakRole.cs
+++ b/SmartTutor/KeycloakAuth/AllowedKeycloakRole.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using Microsoft.AspNetCore.Authorization;
+
+namespace SmartTutor.KeycloakAuth
+{
+    public class AllowedKeycloakRole : IAuthorizationRequirement
+    {
+        public AllowedKeycloakRole(string role)
+        {
+            AllowedRole = role;
+        }
+
+        public string AllowedRole { get; }
+    }
+}
--- a/SmartTutor/Startup.cs
+++ b/SmartTutor/Startup.cs
@@ -20,7 +20,7 @@ using SmartTutor.ProgressModel;
 using SmartTutor.ProgressModel.Feedback.Repository;
 using SmartTutor.ProgressModel.Progress.Repository;
 using SmartTutor.ProgressModel.Submissions.Repository;
-using SmartTutor.KeycloakAuth;
+using SmartTutor.Controllers.KeycloakAuth;
 
 namespace SmartTutor
 {
--- a/SmartTutor/Startup.cs
+++ b/SmartTutor/Startup.cs
@@ -63,6 +63,22 @@ namespace SmartTutor
             
             services.AddScoped<IInstructor, VARKRecommender>();
 
+            AuthenticationConfig(services);
+            AuthorizationConfig(services);
+        }
+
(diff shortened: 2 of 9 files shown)


SmartTutor/KeycloakAuth/AllowedKeycloakRole.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace SmartTutor.KeycloakAuth
+{
+    public class AllowedKeycloakRole : IAuthorizationRequirement
+    {
+        public AllowedKeycloakRole(string role)
+        {
+            AllowedRole = role;
+        }
+
+        public string AllowedRole { get; }
```
Luburic, line 13: Will this be the code from which we build our RBAC model? We should explore how RBAC ties in with keycloak - does the role-permissions mapping need to be handled on our side or on Keycloak's side.

SmartTutor/KeycloakAuth/AllowedKeycloakRole.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Microsoft.AspNetCore.Authorization;
+
+namespace SmartTutor.KeycloakAuth
+{
+    public class AllowedKeycloakRole : IAuthorizationRequirement
```
Luburic, line 6: Do we need the prefix "Allowed" for these concepts? Are there any other kinds of roles?

SmartTutor/Startup.cs
```diff
@@ -47,6 +53,42 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<ITraineeRepository, TraineeDatabaseRepository>();
             services.AddScoped<IFeedbackRepository, FeedbackDatabaseRepository>();
             services.AddScoped<IRecommender, KnowledgeBasedRecommender>();
+
+            services.AddAuthentication(options =>
```
Luburic, line 57: Can we extract this setup into a private function (e.g., ConfigureAuthentication)?

Luburic (changes requested):
A few minor changes to address.

SmartTutor/KeycloakAuth/AllowedKeycloakRole.cs
lukaDoric: We don't have any RBAC model in backend for now. Role is claimed from JWT that is gathered from Keycloak (Role system is managed on Keycloak side). On the backend side we just configure policy for the role that comes from Keycloak.

Luburic (commented):
Looks mostly good, but there are two things remaining:
- Resolve conflict
- Move KeycloakAuth to the Controllers package.
After that you can merge.

PR #42, opened 2021-04-29
Annotations sanity check

simsimkic:
Sanity check should check:

- [x]  Consistency of a single annotator: did annotator gave the same severity values for similar instances, where similarity is observed through metric values
- [x]  Consistency between all annotators: did all annotators gave the same severity value for similar instances

Consistency is calculated using MANOVA test (multivariate ANOVA).


CodeModel/CodeParsers/CSharp/CSharpCodeParser.cs
```diff
@@ -259,23 +260,27 @@ private List<CaDETVariable> LinkMethodVariables(List<CaDETClass> classes, CaDETM
...
             if (foundType != null) return foundType;
-
             foundType = GetArrayType(classes, fullType);
             return foundType;
         }
         
         private List<CaDETClass> GetCollectionTypes(List<CaDETClass> classes, string type)
         {
             List<CaDETClass> collectionTypes = new List<CaDETClass>();
+            if (type == null) return collectionTypes;
```
Luburic, line 283: A lot of null checks everywhere. Is it possible to centralize this validation check instead of repeating it in every method?

DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
```diff
@@ -91,6 +91,12 @@ private int GetAnnotationFromMostExperiencedAnnotator()
             return Annotations.OrderBy(a => a.Annotator.Ranking).First().Severity;
         }
 
+        public bool IsAnnotatedBy(int annotatorId)
+        {
+            if (Annotations.Count(a => a.Annotator.Id == annotatorId) == 0) return false;
```
Luburic, line 96: return Annotations.Any(a => a.Annotator.Id == annotatorId);

DataSetExplorer/DataSetSerializer/BetweenAnnotatorsSanityCheckExporter.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using CodeModel.CaDETModel.CodeItems;
+using DataSetExplorer.DataSetBuilder.Model;
+using OfficeOpenXml;
+
+namespace DataSetExplorer.DataSetSerializer
+{
+    class BetweenAnnotatorsSanityCheckExporter
```
Luburic, line 10: This name is interesting :)

What do the terms "BetweenAnnotators" and "SanityCheck" mean? We need terminology that can be understood by anyone familiar with the domain - I think that Koča used sanity check as a generic term, and betweenannotators isn't really a valid term. Ask the domain expert for a more formal definition to expand our ubiqutious language.

DataSetExplorer/DataSetSerializer/SingleAnnotatorSanityCheckExporter.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace DataSetExplorer.DataSetSerializer
+{
+    class SingleAnnotatorSanityCheckExporter
+    {
+        private readonly string _templatePath = "../../../DataSetSerializer/Template/Single_Annotator_Sanity_Check_Template.xlsx";
+        private readonly string _exportPath;
+        private ExcelPackage _excelFile;
+        private ExcelWorksheet _sheet;
+
+        public SingleAnnotatorSanityCheckExporter(string exportPath)
```
Luburic, line 18: This class shares most of the code with the previous one. Can this be a single class that offers two public methods?

DataSetExplorer/DataSetSerializer/BetweenAnnotatorsSanityCheckExporter.cs
```diff
@@ -0,0 +1,66 @@
...
+            string fileName, int severity)
+        {
+            ExcelPackage.LicenseContext = LicenseContext.NonCommercial;
+            _excelFile = new ExcelPackage(new FileInfo(_templatePath));
+            _sheet = _excelFile.Workbook.Worksheets[0];
+            PopulateTemplate(dataSetInstances, severity);
+            Serialize(fileName + severity);
+        }
+
+        private void PopulateTemplate(List<Tuple<DataSetInstance, Dictionary<CaDETMetric, double>>> instances, int severity)
```
Luburic, line 32: Why is severity important here?

DataSetExplorer/DataSetSerializer/SingleAnnotatorSanityCheckExporter.cs
```diff
@@ -0,0 +1,60 @@
...
+            string fileName, int annotatorId)
+        {
+            ExcelPackage.LicenseContext = LicenseContext.NonCommercial;
+            _excelFile = new ExcelPackage(new FileInfo(_templatePath));
+            _sheet = _excelFile.Workbook.Worksheets[0];
+            PopulateTemplate(dataSetInstances, annotatorId);
+            Serialize(fileName + annotatorId);
+        }
+
+        private void PopulateTemplate(List<Tuple<DataSetInstance, Dictionary<CaDETMetric, double>>> instances, int annotatorId)
```
Luburic, line 33: This tuple might be a new domain concept. When generating a dataset we might want to include these metrics, so we should think about expanding our domain model.

DataSetExplorer/ManovaTest/ManovaTest.cs
```diff
@@ -0,0 +1,44 @@
...
+        
+        public void SetupTestArguments(string annotatedInstancesFile, List<CaDETMetric> metrics, string independentVariable)
+        {
+            _annotatedInstancesFile = annotatedInstancesFile;
+            _metrics = "";
+            foreach (var metric in metrics)
+            {
+                _metrics += metric + " + ";
+            }
+            _metrics = _metrics[0..^2];
```
Luburic, line 25: ?

Luburic (changes requested):
Comments inline.

DataSetExplorer/ManovaTest/ManovaTest.cs
simsimkic: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/ranges-indexes

In this example, I take from index 0 to length - 2.

Luburic: Another option is to do _metrics = string.join(" + ", metrics).

DataSetExplorer/DataSetSerializer/BetweenAnnotatorsSanityCheckExporter.cs
simsimkic: We run Manova test for each severity, and I generate separate files for those tests.

image

DataSetExplorer/DataSetSerializer/AnnotationConsistencyByMetricsExporter.cs
```diff
@@ -0,0 +1,91 @@
...
+        private readonly string _exportPath;
+        private ExcelPackage _excelFile;
+        private ExcelWorksheet _sheet;
+
+        public AnnotationConsistencyByMetricsExporter(string exportPath)
+        {
+            _exportPath = exportPath;
+        }
+
+        public void ExportAnnotationsFromAnnotator(int annotatorId, List<Tuple<DataSetInstance, Dictionary<CaDETMetric, double>>> dataSetInstances,
```
Luburic, line 24: Now that the DataSetInstance has the Metrics dictionary, do we need this Tuple?

DataSetExplorer/DataSetBuilder/CaDETToDataSetBuilder.cs
```diff
@@ -115,8 +115,12 @@ private int DetermineNumberOfInstances(int totalNumber)
 
         private List<DataSetInstance> CaDETToDataSetClasses(List<CaDETClass> cadetClasses)
         {
-            return cadetClasses.Select(c => 
-                new DataSetInstance(c.FullName, GetCodeUrl(c.FullName), _projectAndCommitUrl, SnippetType.Class)).ToList();
+            return cadetClasses.Select(c =>
+            {
+                var dataSetInstance = new DataSetInstance(c.FullName, GetCodeUrl(c.FullName), _projectAndCommitUrl, SnippetType.Class);
```
Luburic, line 120: Could we avoid the anonymous function with an expanded constructor?

Luburic (commented):
A few minor things to address, as well as the results of the CodeFactor, and you can merge.

DataSetExplorer/DataSetBuilder/CaDETToDataSetBuilder.cs
simsimkic: Do you think we should remove the select method? As far as I understand, we need it to transform all classes to data set instances. Maybe I didn't understand the comment because I'm not completely familiar with C#. :) Do you mean:

return cadetClasses.Select(c =>
    {
        return new DataSetInstance(c.FullName, GetCodeUrl(c.FullName), _projectAndCommitUrl, SnippetType.Class)
        {
            MetricFeatures = _cadetProject.GetMetricsForCodeSnippet(c.FullName)
        };
    }).ToList();

Luburic: I was thinking more along the lines of:

    return cadetClasses.Select(c => return new DataSetInstance(
            c.FullName, GetCodeUrl(c.FullName), _projectAndCommitUrl, SnippetType.Class, _cadetProject.GetMetricsForCodeSnippet(c.FullName)
        ).ToList()

Luburic (approved):
Once you finalize the last comment feel free to merge.

PR #45, opened 2021-05-05
Project strategy

Ana00000:
- New strategy based on multiple fulfillment strategies for snippets of code


SmartTutor/ContentModel/LearningObjects/Challenges/FulfillmentStrategy/ProjectChecker/ProjectAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                var result = strategy.EvaluateSubmission(GetAllClassesFromSnippets(solutionAttempt));
+                challengeHints.MergeHints(result);
+            }
+            return challengeHints;
+        }
+
+        public override List<ChallengeHint> GetAllHints()
+        {
+            return SnippetApplicableStrategies.SelectMany(sas => sas.Value).SelectMany(cfs => cfs.GetAllHints().Where(h => h != null)).ToList();
```
Luburic, line 31: We can remove the Where(h => h != null) as the Challenge already performs this check.

SmartTutor/ContentModel/LearningObjects/Challenges/FulfillmentStrategy/ProjectChecker/ProjectAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
+﻿using CodeModel.CaDETModel.CodeItems;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SmartTutor.ContentModel.LearningObjects.Challenges.FulfillmentStrategy.ProjectChecker
+{
+    public class ProjectAnalyzer : ChallengeFulfillmentStrategy
+    {
+        public Dictionary<string, List<ChallengeFulfillmentStrategy>> SnippetApplicableStrategies { get; private set; }
```
Luburic, line 9: StrategiesApplicableToSnippet would be more appropriate.

SmartTutor/ContentModel/LearningObjects/Challenges/FulfillmentStrategy/ProjectChecker/ProjectAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            SnippetApplicableStrategies = snippetApplicableStrategies;
+        }
+
+        public override HintDirectory EvaluateSubmission(List<CaDETClass> solutionAttempt)
+        {
+            var challengeHints = new HintDirectory();
+            foreach (var strategy in SnippetApplicableStrategies.SelectMany(sas => sas.Value))
+            {
+                var result = strategy.EvaluateSubmission(GetAllClassesFromSnippets(solutionAttempt));
```
Luburic, line 23: If I'm not mistaken, this code takes all the strategies and applies them to all the CaDETClasses, asking each one if it contains the suitable snippet below. This has a major readability issue - the EvaluateSubmission is a high-level method that is read first and that should give us an idea of what the class is doing without requiring us to read the details. When I'm reading this method I see that all the strategies from the dictionary are applied to all the classes (from snippets?).

Only when I go into the details do I see that the key is used (and the below functions are confusing to grasp - at least more than they should be).

A more straightforward flow of logic would be:

- For each code snippet for which we have strategies:
    - Find the code item (either CaDETClass or CaDETMember) based on the snippet id. Throw an exception if there is no such item.
    - Apply the strategies related to that snippet.
    - Merge the obtained hints.

Changing this to code, we would have something along the lines of:


    foreach (var codeSnippetId in SnippetApplicableStrategies.Keys) {
        var codeSnippet = FindCodeSnippet(codeSnippetId, solutionAttempt); //if snippedId contains () looks at method or constructor signatures and returns empty class with that member, otherwise classes. Throws exception if not found.
        foreach(var strategy in SnippetApplicableStrategies[codeSnippetId]) {
            challengeHints.MergeHints(strategy.EvaluateSubmission(codeSnippet);
        }
    }

SmartTutor/ContentModel/LearningObjects/Challenges/FulfillmentStrategy/ProjectChecker/ProjectAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
+﻿using CodeModel.CaDETModel.CodeItems;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SmartTutor.ContentModel.LearningObjects.Challenges.FulfillmentStrategy.ProjectChecker
+{
+    public class ProjectAnalyzer : ChallengeFulfillmentStrategy
```
Luburic, line 7: ProjectChecker might be more suitable to adhere to the convention of the other strategies.

SmartTutor/ContentModel/LearningObjects/Challenges/FulfillmentStrategy/ProjectChecker/ProjectChecker.cs
```diff
@@ -0,0 +1,70 @@
...
+                    FullName = fullName,
+                    Name = GetNameOfSnippet(fullName),
+                    Fields = new List<CaDETField>()
+                };
+            }
+
+            throw new KeyNotFoundException("Code snippet id wasn't found.");
+        }
+
+        private string GetFullNameOfSnippet(string codeSnippetId)
```
Luburic, line 58: Do we need the information provided by GetFullNameOfSnippet and GetNameOfSnippet? If not, we should remove this extra code and create a "noname" class with the member.

SmartTutor/Database/SmartTutorContext.cs
```diff
@@ -100,5 +108,21 @@ private static void ConfigureBasicMetricChecker(ModelBuilder modelBuilder)
...
+               .Property<string?>("StrategiesApplicableToSnippetCheckerForeignKey").IsRequired(false);
+
+            modelBuilder.Entity<ProjectChecker>()
+                .Property(pc => pc.StrategiesApplicableToSnippet)
+                .IsRequired()
+                .HasConversion<string[]>(
+                    sats => ((IEnumerable)sats).Cast<object>().Select(x => x.ToString()).ToArray(),
+                    sats => sats == null
+                        ? new Dictionary<string, List<ChallengeFulfillmentStrategy>>()
+                        : JsonConvert.DeserializeObject<Dictionary<string, List<ChallengeFulfillmentStrategy>>>(sats.ToString())
```
Luburic, line 124: Instead of introducing a new dependency (Netwonsoft), we should use the built in JSON serializer (https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?pivots=dotnet-5-0).

In general, we want to depend on the least amount of external code possible. While it may be convenient to plug in a new library with a few clicks there is a price that we pay for maintaining a link to that library. Ideally, we can do our job with the built in functionality, as is the case with the System.Text.Json package.

SmartTutor/SmartTutor.csproj
```diff
@@ -14,6 +14,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
Luburic, line 17: Should be removed.

Luburic (changes requested):
A few minor comments to address.

SmartTutor/ContentModel/LearningObjects/Challenges/FulfillmentStrategy/ProjectChecker/ProjectChecker.cs
Ana00000: In further evaluation of the code, it's necessary to put at least " " for those attributes (or execution will break), so that's why I added them. If it isn't of great value I can remove those two methods and set those names to " ".

SmartTutor/Database/SmartTutorContext.cs
```diff
@@ -100,5 +110,24 @@ private static void ConfigureBasicMetricChecker(ModelBuilder modelBuilder)
...
+
+        private static void ConfigureProjectChecker(ModelBuilder modelBuilder)
+        {
+            modelBuilder.Entity<ChallengeFulfillmentStrategy>()
+               .Property<string>("StrategiesApplicableToSnippetCheckerForeignKey").IsRequired(false);
+
+            modelBuilder.Entity<ProjectChecker>()
+                .Property(pc => pc.StrategiesApplicableToSnippet)
+                .IsRequired()
+                .HasConversion(
```
Luburic, line 122: We should discuss this.

SmartTutorTests/DataFactories/SemanticProjectCheckerChallengeTestData.cs
```diff
@@ -0,0 +1,231 @@
+﻿namespace SmartTutor.Tests.DataFactories
+{
+    internal class SemanticProjectCheckerChallengeTestData
+    {
+        public static string[] GetFourPassingClasses()
+        {
+            return new[]
+           {
+@"
```
Luburic, line 9: This should be exported into a .txt file that is labeled to avoid compilation (we can discuss this today).

SmartTutorTests/Unit/ProjectCheckerTests.cs
```diff
@@ -0,0 +1,229 @@
...
+using SmartTutor.ContentModel.LearningObjects.Challenges.FulfillmentStrategy;
+using SmartTutor.ContentModel.LearningObjects.Challenges.FulfillmentStrategy.MetricChecker;
+using SmartTutor.ContentModel.LearningObjects.Challenges.FulfillmentStrategy.NameChecker;
+using SmartTutor.ContentModel.LearningObjects.Challenges.FulfillmentStrategy.ProjectChecker;
+using SmartTutor.Tests.DataFactories;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace SmartTutor.Tests.Unit
```
Luburic, line 12: We should discuss this as well.

Luburic (commented):
Generally looks good, but we have a few things to discuss.

Luburic (changes requested):
Requires further work, but we will pause it for now.

Luburic:
For now we will pause development on this feature and close the PR. Once a better serialization strategy is designed we can reopen the PR.

PR #46, opened 2021-05-08
feat: Checks if question or challenge is in any of learners courses

VladimirBudjen:


SmartTutorTests/Integration/SubmissionTests.cs
```diff
@@ -135,7 +153,8 @@ public void Accepts_arrange_task_submission_and_produces_correct_evaluation(Arra
                 true
             }
         };
-
+
```
VladimirBudjen, line 156: I will delete this line in next commit. 😄

SmartTutorTests/Unit/SubmissionTests.cs
```diff
@@ -68,22 +99,24 @@ public void Evaluates_answer_submission(QuestionSubmission submission, List<bool
             correctness.ShouldBe(expectedCorrectness);
         }
 
+        private static readonly Submission Submission = new Submission(1, 1, true, DateTime.Now);
```
Luburic, line 102: Remove this. Submission should be an abstract class as it does not make sense as a standalone object (submission of what?). As a side note, we should avoid sharing state between tests, even when it's immutable - if this test suite grows and more tests start to depend on this shared field, what happens when one test changes it?

Why do you need these fields in the test data? If it turns out we don't, don't forget to remove the additional parameter you added to the QuestionSubmission constructor, and the two constructors you added to Submission.

SmartTutor/ProgressModel/Submissions/Submission.cs
```diff
@@ -13,5 +13,25 @@ public void MarkCorrect()
...
             IsCorrect = true;
         }
+
+        protected Submission()
+        {
+        }
+
+        public Submission(int id, int learnerId, bool isCorrect, DateTime timeStamp)
+        {
+            Id = id;
```
Luburic, line 23: Look at my comment in the SubmissionTests class, then come back here.

Luburic (changes requested):
A few minor things to address, including the CodeFactor tip.

Luburic (approved):
Looks good.

PR #50, opened 2021-05-24
ICMBC cohesion metric

Luburic:
Opening PR for easier review.


CodeModel/CodeParsers/CSharp/ICBMCGraph.cs
```diff
@@ -0,0 +1,336 @@
...
+                   m.Metrics[CaDETMetric.MELOC] > 1; // && m.Modifiers.Contains(new CaDETModifier("public"));
+        }
+
+        private int[,] InitializeMatrix(List<CaDETMember> normalMethods, List<CaDETField> fields,
+            List<CaDETMember> fieldDefiningAccessors)
+        {
+            int[,] matrix = new int[normalMethods.Count, fields.Count + fieldDefiningAccessors.Count];
+            for (var i = 0; i < normalMethods.Count; i++)
+            {
+                for (var j = 0; j < fields.Count; j++)
```
Luburic, line 48: There are 4-5 places where you have to check if we are working with fields or methods, and the code is mostly the same for both cases... This seems like a smell, but I'm not sure how we could refactor this. We might know more once we start transforming this to include indirect field access through private methods etc.

Luburic (commented):
I've updated the branch to include the recent master changes. See my inline comments for now and assign me for review when you push the changes for the left and right subgraph and the renaming to avoid technical terminology (e.g., matrix).

balsa-sarenac:
Closed in favour of: https://github.com/Clean-CaDET/platform/pull/74

PR #51, opened 2021-05-25
Check metrics significance in annotations

simsimkic:
Runs the ANOVA test and checks if the code metrics are significant for given annotations (severities).


DataSetExplorer/Program.cs
```diff
@@ -38,15 +38,42 @@ static void Main(string[] args)
...
+            var exporter = new AnnotationConsistencyByMetricsExporter("D:/ccadet/annotations/sanity_check/anova/Output/");
+            var anovaTest = new AnnotationConsistencyTests.AnovaTest();
+
+            var enumerator = instancesGroupedBySmells.GetEnumerator();
+            while (enumerator.MoveNext())
+            {
+                var codeSmellGroup = enumerator.Current;
+                var codeSmell = codeSmellGroup.Key.Replace(" ", "_");
+
+                exporter.ExportAnnotationsFromAnnotator(annotatorId, codeSmellGroup.ToList(),
```
Luburic, line 58: Here you have a minor case of shotgun surgery - if I want to change the path where I want to place the results, I need to change LOC 49, LOC 64 and possibly the body of the ExportAnnotationsFromAnnotator (e.g., if I want to remove sanity check).

Furthermore, if the exporter changes the way it creates the path for storing an excel, this will break the anovaTest.

Come up with a design that centralizes the base path and relative path information.

Luburic (changes requested):
Comments inline.

DataSetExplorer/AnnotationConsistencyTests/AnovaTest.cs
```diff
@@ -0,0 +1,52 @@
...
+using DataSetExplorer.DataSetBuilder.Model;
+using DataSetExplorer.DataSetSerializer;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+
+namespace DataSetExplorer.AnnotationConsistencyTests
+{
+    class AnovaTest
```
Luburic, line 11: internal?

DataSetExplorer/Program.cs
```diff
@@ -81,35 +92,25 @@ private static ListDictionary GetAnnotatedProjects()
...
         {
             var instancesGroupedBySmells = GetAnnotatedInstancesGroupedBySmells(annotatorId);
+            var manovaTest = new AnnotationConsistencyTests.ManovaTest();
 
-            var exporter = new AnnotationConsistencyByMetricsExporter("D:/ccadet/annotations/sanity_check/Output/");
-            var manovaTest = new ManovaTest.ManovaTest();
-
-            var enumerator = instancesGroupedBySmells.GetEnumerator();
-            while (enumerator.MoveNext())
+            foreach (var codeSmellGroup in instancesGroupedBySmells)
```
Luburic, line 97: You have 3 repetitions of this code, where only the final command is different. Can you come up with an alternative design?

This Program.cs class will require major refactoring anyway so it is not that important that we solve it right now, but contemplate this issue a bit to see if you can come up with a better design.

Luburic (commented):
Will suffice for now, a few minor comments to consider but you can merge these changes for now. We will have a PR or two dedicated to refactoring this Program.cs soon.

DataSetExplorer/AnnotationConsistencyTests/ManovaTest.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        private string ExportAnnotatorsForSeverity(int severity, List<DataSetInstance> instances, string codeSmell)
+        {
+            var exportedAnnotatorsFile = "SanityCheck_" + codeSmell + "_Severity_" + severity;
+            var exporter = new AnnotationConsistencyByMetricsExporter(_annotatedInstancesFolderPath);
+            exporter.ExportAnnotatorsForSeverity(severity, instances, exportedAnnotatorsFile);
+            return exportedAnnotatorsFile;
+        }
+
+        public void TestForSingleAnnotator(int annotatorId, List<DataSetInstance> instances, string codeSmell, List<CaDETMetric> metrics)
```
Luburic, line 61: The problem with the current implementation is that the client of ManovaTest (Program) needs to know all the details of this class, to select the appropriate delegate. It is tightly coupled to this object. There are a few things to consider here:

- What is the difference in code between this method and `TestForMultipleAnnotators` apart from the first line and the third argument of the second line? Also examine the RunTest of AnovaTest.cs, as well as the similarities in the classes' fields.
- Maybe it would be helpful to look at the big picture - what do the classes Manova and Anova test do in a broader application - what is their purpose? Are they implementations of the same interface that holds a domain significant name (e.g., AnnotatorConsistencyChecker)?
- In such a system, what are the data flows? How can we hide the details (including which delegate to choose) from the client of these Tests? Introducing (at least imaginary) interfaces helps us understand what the role of the code implementing the interface is, and how to encapsulate all the details apart from the (often single) meaningful function call.

My advice is to think about this problem because it is a good one - sketch some class and communication diagrams of this larger system, think about the different moving parts, and how you could make the client unaware of which statistical test it is working with. Examine the easiest way to support a third type of statistical test. It takes time, but this is where your design skills improve.

Luburic (changes requested):

DataSetExplorer/AnnotationConsistencyTests/AnovaTest.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            return Test(annotatorId, instancesGroupedBySmells, TestCodeSmellForAnnotator);
+        }
+
+        public Result<Dictionary<string, Dictionary<string, string>>> TestBetweenAnnotators(int severity, IEnumerable<IGrouping<string, DataSetInstance>> instancesGroupedBySmells)
+        {
+            return Test(severity, instancesGroupedBySmells, TestCodeSmellBetweenAnnotators);
+        }
+
+        private Result<Dictionary<string, Dictionary<string, string>>> Test(int id, IEnumerable<IGrouping<string, DataSetInstance>> instancesGroupedBySmells, TestCodeSmellDelegate testCodeSmell)
```
Luburic, line 33: Instead of Dictionary<string, Dictionary<string, string>> can we introduce a meaningful class (i.e., value object)?

If not for the outer dictionary, for the inner dictionary?

If we can't come up with a meaningful concept, we should pay attention to this data structure and see how it changes.

Luburic (approved):
One comment to think about. Take 15 minutes to think about the issue and see if there is a good name for the Dictionary<string, string>. If nothing comes up, merge.

PR #53, opened 2021-06-02
Data set files with annotations, metrics and heuristics

simsimkic:
Generate three types of files for a data set:

1. A file containing basic information about instances and annotations (individual annotations from annotators and final annotation)
2. A file containing the instance id and metrics' values
3. A file containing the instance id and heuristics' values


DataSetExplorer/DataSetSerializer/CompleteDataSetExporter.cs
```diff
@@ -10,28 +10,38 @@ namespace DataSetExplorer.DataSetSerializer
...
 
         public CompleteDataSetExporter(string exportPath)
         {
             _exportPath = exportPath;
         }
 
-        public void Export(List<DataSetInstance> dataSetInstances, string fileName)
+        public void Export(List<DataSetInstance> dataSetInstances, string smell, string fileName)
+        {
+            ExportDataSetWithAnnotations(dataSetInstances, fileName);
```
Luburic, line 28: Maybe each of these segments deserves a separate class. I would say "no" at this moment, but if the complexity of this class continues to grow (and surpasses roughly 250 LOC) I would consider splitting the class, especially if there is no or very little shared logic.

On a related matter, the `Export`, `ExportDataSetWithAnnotations`, `ExportDataSetWithMetrics`, and `ExportDataSetWithHeuristics `roughly have the same API. If we go through with the extractions, it might make sense to introduce an interface and have the `CompleteDataSetExporter` implement it and maintain a list of `IDatasetExporter`s.

Luburic (changes requested):
Minor improvements required.

DataSetExplorer/DataSetSerializer/CompleteDataSetExporter.cs
simsimkic: I was thinking the same. I will keep this in mind.

Luburic (approved):

PR #55, opened 2021-06-04
Implemented missing metrics for GC detection

katarina-glorija:
LCOM3 - Lack of cohesion based on https://core.ac.uk/download/pdf/231161496.pdf
LCOM4 - Lack of cohesion based on http://www.isys.uni-klu.ac.at/PDF/1995-0043-MHBM.pdf
DIT -  Depth inheritance hierarchy based on https://objectscriptquality.com/docs/metrics/depth-inheritance-tree
DCC -  Direct class coupling based on http://www.theijes.com/papers/v5-i6/C05014019.pdf

### DIFF
--- a/CodeModel.Tests/DataFactories/CodeFactory.cs
+++ b/CodeModel.Tests/DataFactories/CodeFactory.cs
@@ -767,6 +767,48 @@ namespace CodeModel.Tests.DataFactories
             };
         }
 
+        public IEnumerable<string> GetCouplingClasses()
+        {
+            return new[]
+            {
+                @"
+            namespace Test.Data
+            {
+                public class Test1
+                {
+                    public int _param1 { get; set; }
+                    public int _param2 { get; set; }
+
+                    public Test1(int param1, int param2)
+                    {
+                        _param1 = param1;
+                        _param2 = param2;
+                    }
+                }
+            }",
+                @"
+            namespace Test.Data
+            {
+                public class Test2
+                {
+                    public Test1 _test;
+                }
+            }",
+                @"
+            namespace Test.Data
+            {
+                public class Test3
+                {
+                    public Test2 _test2;
+
+                    public Test1 method(Test1 test){
+                        return test;
+                    }
+                }
+            }"
+            };
+        }
+
         public IEnumerable<string> GetClassesFromDifferentNamespace()
         {
             return new[]
--- a/CodeModel.Tests/DataFactories/CodeFactory.cs
+++ b/CodeModel.Tests/DataFactories/CodeFactory.cs
@@ -4,7 +4,6 @@ namespace CodeModel.Tests.DataFactories
 {
     public class CodeFactory
     {
-
         public IEnumerable<string> readClassFromFile(string path)
         {
             return new[] { System.IO.File.ReadAllText(path) };
--- a/CodeModel.Tests/DataFactories/CodeFactory.cs
+++ b/CodeModel.Tests/DataFactories/CodeFactory.cs
@@ -4,6 +4,12 @@ namespace CodeModel.Tests.DataFactories
 {
     public class CodeFactory
     {
+
+        public IEnumerable<string> readClassFromFile(string path)
+        {
+            return new[] { System.IO.File.ReadAllText(path) };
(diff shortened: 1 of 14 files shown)


CodeModel.Tests/Unit/MetricCalculationTests.cs
```diff
@@ -69,6 +69,32 @@ public void Calculates_lack_of_cohesion()
...
+        public void Calculates_lack_of_cohesion_3()
+        {
+            CodeModelFactory factory = new CodeModelFactory();
+
+            List<CaDETClass> classes = factory.CreateProject(_testDataFactory.GetCohesionClasses()).Classes;
+
+            var dateRange = classes.Find(c => c.Name.Equals("DateRange"));
+            var doctor = classes.Find(c => c.Name.Equals("Doctor"));
+            dateRange.Metrics[CaDETMetric.LCOM3].ShouldBe(0);
+            doctor.Metrics[CaDETMetric.LCOM3].ShouldBe(1.125);
```
Luburic, line 82: Did you take this test data and calculated values from the paper? Ideally we would have a test suite that matches the examples provided by the authors.

CodeModel.Tests/DataFactories/CodeFactory.cs
```diff
@@ -767,6 +767,48 @@ IEnumerator IEnumerable.GetEnumerator()
             };
         }
 
+        public IEnumerable<string> GetCouplingClasses()
```
Luburic, line 770: This class needs to die. We'll schedule a meeting to plot its murder.

CodeModel/CaDETModel/CodeItems/CaDETClass.cs
```diff
@@ -45,6 +45,20 @@ public List<CaDETClass> GetMethodLinkedVariableTypes()
...
+        {
+            List<CaDETParameter> parameters = Members.SelectMany(m => m.Params).ToList();
+            return parameters.Select(p => p.Type)
+                   .Where(v => v.LinkedTypes != null)
+                   .SelectMany(v => v.LinkedTypes).ToList();
+        }
+
+        public List<CaDETClass> GetMethodInvocationsTypes()
+        {
+            List<CaDETMember> invokedMethods = Members.SelectMany(m => m.InvokedMethods).Where(m => m.Type is CaDETMemberType.Method).ToList();
```
Luburic, line 58: Invoked methods are already only methods (which is why we don't call them invoked members).

CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -69,6 +74,55 @@ private double GetLackOfCohesionOfMethods(CaDETClass parsedClass)
...
+                methodFieldAccess += CountOwnFieldAndAccessorAccessed(parsedClass, method);
+            }
+            return Math.Round((numberOfMethods - methodFieldAccess/numberOfAttributes) / (numberOfMethods - 1), 3);
+        }
+
+        /// <summary>
+        /// LCOM - Lack of Cohesion of Methods
+        /// DOI: 10.1145/2723742.2723753
+        /// </summary>
+        private double GetLackOfCohesionOfMethods4(CaDETClass parsedClass)
```
Luburic, line 101: Given the complexity of this calculation we need a lot of test data. Search online if the papers don't give enough examples and have 5+ test cases for each metric.

Luburic (changes requested):
A few minor changes, plus a task to enhance the test suite with (many) more test cases gathered from online resources. Given the nature of metrics and the issues that arise from their (mis)implementation, we need to significantly expand our test suite.

CodeModel.Tests/Unit/MetricCalculationTests.cs
katarina-glorija: I calculated them manually. Unfortunately, the paper doesn't provide any example.

Luburic: Then look to the web for inspiration.

Luburic (approved):
Looks good.

PR #59, opened 2021-07-01
Service logic in MVP DataSet Explorer

simsimkic:
Following image illustrates extracted services and their use cases:

image


DataSetExplorer/DataSetSerializer/IDataSetExporter.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace DataSetExplorer
+{
+    public interface IDataSetExporter
```
Luburic, line 3: Maybe we should include the keyword Service for all of these interfaces (including the earlier Creator interface I made).

I can see how this interface can be used for a low-level technical implementation (e.g., ExcelExporter), but our aim is to have a full use case implementation.

For domain services we don't want to use the keyword Service, so the IFullDataSetBuilder should remain the same. This will be helpful to differentiate domain and application services.

DataSetExplorer/DataSetSerializer/IDataSetExporter.cs
```diff
@@ -0,0 +1,7 @@
...
+    {
+        public void Export(string outputPath);
```
Luburic, line 5: Should return Result<string> so that we can have an error if, for example, the outputPath targets a folder for which the app doesn't have permission to access.

DataSetExplorer/DataSetBuilder/IDataSetAnalyzer.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace DataSetExplorer
+{
+    interface IDataSetAnalyzer
```
Luburic, line 3: See comment at "IDataSetExporter".

DataSetExplorer/DataSetBuilder/FullDataSetBuilderService.cs
```diff
@@ -0,0 +1,70 @@
...
+            return annotatedInstances.Select(i => {
+                i.MetricFeatures = project.GetMetricsForCodeSnippet(i.CodeSnippetId);
+                return i;
+            }).ToList();
+        }
+
+        private ListDictionary GetAnnotatedProjects()
+        {
+            ListDictionary projects = new ListDictionary(); // local repository path, annotations folder path
+            projects.Add("D:/ccadet/annotations/repos/BurningKnight", "D:/ccadet/annotations/annotated/BurningKnight");
```
Luburic, line 44: We can't have this hardcoded.

DataSetExplorer/DataSetBuilder/FullDataSetBuilderService.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        private List<DataSetInstance> FillInstancesWithMetrics(List<DataSetInstance> annotatedInstances, CaDETProject project)
+        {
+            return annotatedInstances.Select(i => {
+                i.MetricFeatures = project.GetMetricsForCodeSnippet(i.CodeSnippetId);
+                return i;
+            }).ToList();
+        }
+
+        private ListDictionary GetAnnotatedProjects()
```
Luburic, line 41: Why not a list or set?

DataSetExplorer/DataSetBuilder/FullDataSetBuilderService.cs
```diff
@@ -0,0 +1,70 @@
...
+            projects.Add("D:/ccadet/annotations/repos/jellyfin", "D:/ccadet/annotations/annotated/Jellyfin");
+            projects.Add("D:/ccadet/annotations/repos/OpenRA", "D:/ccadet/annotations/annotated/OpenRA");
+            projects.Add("D:/ccadet/annotations/repos/ShareX", "D:/ccadet/annotations/annotated/ShareX");
+            projects.Add("D:/ccadet/annotations/repos/ShopifySharp", "D:/ccadet/annotations/annotated/ShopifySharp");
+            projects.Add("D:/ccadet/annotations/repos/MonoGame", "D:/ccadet/annotations/annotated/MonoGame");
+            projects.Add("D:/ccadet/annotations/repos/Osu", "D:/ccadet/annotations/annotated/Osu");
+            return projects;
+        }
+
+        private void LoadAnnotators(ref List<DataSetInstance> annotatedInstances)
```
Luburic, line 55: ref?

DataSetExplorer/DataSetBuilder/FullDataSetBuilderService.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            List<Annotator> annotators = new List<Annotator>()
+            {
+                new Annotator(1, 6, 1),
```
Luburic, line 59: Hardcoded? I realize this is not the focus of the PR, but you should leave a TODO comment to ensure you don't forget to resolve these issues later.

DataSetExplorer/DataSetBuilder/FullDataSetBuilderService.cs
```diff
@@ -0,0 +1,70 @@
...
+using CodeModel.CaDETModel;
+using DataSetExplorer.DataSetBuilder.Model;
+using DataSetExplorer.DataSetSerializer;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Linq;
+
+namespace DataSetExplorer
+{
+    class FullDataSetBuilderService : IFullDataSetBuilder
```
Luburic, line 11: We should also drop the Service keyword because this is a domain service, to avoid confusion with application services. Actually, we don't need the interface, as the domain service is part of the core layer (in the same way we don't need interfaces for entities and value objects).

We should drop the Builder keyword, as that signals the "Builder" pattern, and this class is not that. I would rename it to FullDataSetFactory, as it's closer to that pattern.

DataSetExplorer/DataSetBuilder/DataSetAnalysisService.cs
```diff
@@ -0,0 +1,28 @@
...
+    class DataSetAnalysisService : IDataSetAnalyzer
+    {
+        public void FindInstancesWithAllDisagreeingAnnotationsUseCase(string dataSetPath, string outputPath)
+        {
+            var dataset = LoadDataSet(dataSetPath);
+            var exporter = new TextFileExporter(outputPath);
+            exporter.ExportInstancesWithAnnotatorId(dataset.GetInstancesWithAllDisagreeingAnnotations());
+        }
+
+        public void FindInstancesRequiringAdditionalAnnotationUseCase(string dataSetPath, string outputPath)
```
Luburic, line 15: Return Result<string> and if you will append UseCase do so consistently (e.g., in DataSetAnalyzer/Exporter).

DataSetExplorer/AnnotationConsistency/IAnnotationConsistencyChecker.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DataSetExplorer
+{
+    interface IAnnotationConsistencyChecker
+    {
+        public void CheckMetricsSignificanceBetweenAnnotatorsForSeverity(int severity);
```
Luburic, line 5: Result<string> or event Result<more-complex-object>.

This will be especially important for the Web, but we should put everything on good footing for the MVP so that we may easily swipe out the Console UI.

DataSetExplorer/Program.cs
```diff
@@ -1,159 +1,26 @@
...
-using CodeModel;
-using CodeModel.CaDETModel;
-using CodeModel.CaDETModel.CodeItems;
-using DataSetExplorer.AnnotationConsistencyTests;
-using DataSetExplorer.RepositoryAdapters;
+﻿using DataSetExplorer.RepositoryAdapters;
 
 namespace DataSetExplorer
 {
     class Program
```
Luburic, line 5: The Program should be responsible for initializing the ConsoleUI and injecting it with the needed dependencies. You can look at https://github.com/Clean-CaDET/platform/blob/master/SmartTutor/Program.cs for inspiration.

Essentially, the Main should be a single line of code "CreateConsoleUI()", where that private method instantiates the ConsoleUI class and runs it. Later, we will add the Startup for the WebUI and simply invoke a different function (similar to the URL above) in the Program based on input params.

Luburic (changes requested):
Several things to improve. You can ignore the comments regarding hardcoded values as this is something you address in a later PR. However, you should pay attention to the principle of writing TODOs for future work so that you can easily search code for these comments.

DataSetExplorer/Program.cs
simsimkic: I will resolve this in https://github.com/Clean-CaDET/platform/pull/60.

DataSetExplorer/DataSetBuilder/FullDataSetBuilderService.cs
simsimkic: I used a ListDictionary to save two pieces of information for each project (e.g., path to the local repository and path to the folder with annotations).

DataSetExplorer/AnnotationConsistency/IAnnotationConsistencyCheckerService.cs
```diff
@@ -0,0 +1,13 @@
+﻿using FluentResults;
+using System.Collections.Generic;
+
+namespace DataSetExplorer
+{
+    interface IAnnotationConsistencyCheckerService
+    {
+        public Result<Dictionary<string, Dictionary<string, string>>> CheckMetricsSignificanceBetweenAnnotatorsForSeverity(int severity);
```
Luburic, line 8: Maybe call the interface IAnnotationConsistencyService?

DataSetExplorer/DataSetBuilder/DataSetAnalysisService.cs
```diff
@@ -0,0 +1,44 @@
+﻿using DataSetExplorer.DataSetBuilder.Model;
+using DataSetExplorer.DataSetSerializer;
+using FluentResults;
+using System;
+
+namespace DataSetExplorer
+{
+    class DataSetAnalysisService : IDataSetAnalyzerService
```
Luburic, line 8: Is it Analysis or Analyzer (looking at the class and interface names).

DataSetExplorer/DataSetBuilder/DataSetAnalysisService.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        public Result<string> FindInstancesWithAllDisagreeingAnnotations(string dataSetPath, string outputPath)
+        {
+            try
+            {
+                var dataset = LoadDataSet(dataSetPath);
+                var exporter = new TextFileExporter(outputPath);
+                exporter.ExportInstancesWithAnnotatorId(dataset.GetInstancesWithAllDisagreeingAnnotations());
+                return Result.Ok("Instances with disagreeing annotations exported: " + outputPath);
+            } catch (Exception e)
```
Luburic, line 18: We should avoid catching the generic Exception. The more specific the better.

DataSetExplorer/DataSetBuilder/DataSetAnalysisService.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public Result<string> FindInstancesRequiringAdditionalAnnotation(string dataSetPath, string outputPath)
+        {
+            try
+            {
+                var dataset = LoadDataSet(dataSetPath);
+                var exporter = new TextFileExporter(outputPath);
+                exporter.ExportInstancesWithAnnotatorId(dataset.GetInsufficientlyAnnotatedInstances());
+                return Result.Ok("Instances requiring additional annotation exported: " + outputPath);
+            } catch (Exception e)
```
Luburic, line 32: Same as above.

DataSetExplorer/DataSetSerializer/DataSetExportationService.cs
```diff
@@ -0,0 +1,34 @@
...
+            try
+            {
+                var instancesGroupedBySmells = _fullDataSetFactory.GetAnnotatedInstancesGroupedBySmells(annotatorId: null);
+                var exporter = new CompleteDataSetExporter(outputPath);
+                foreach (var codeSmellGroup in instancesGroupedBySmells)
+                {
+                    exporter.Export(codeSmellGroup.ToList(), codeSmellGroup.Key, "DataSet_" + codeSmellGroup.Key);
+                }
+                return Result.Ok("Data set exported: " + outputPath);
+            } catch (Exception e)
```
Luburic, line 28: Generic.

DataSetExplorer/AnnotationConsistency/AnnotationConsistencyService.cs
```diff
@@ -0,0 +1,44 @@
...
+        public AnnotationConsistencyService(FullDataSetFactory fullDataSetFactory)
+        {
+            _fullDataSetFactory = fullDataSetFactory;
+        }
+
+        public Result<Dictionary<string, Dictionary<string, string>>> CheckMetricsSignificanceBetweenAnnotatorsForSeverity(int severity)
+        {
+            var instancesGroupedBySmells = _fullDataSetFactory.GetAnnotatedInstancesGroupedBySmells(annotatorId: null);
+            IMetricsSignificanceTester tester = new AnovaTest();
+            return tester.TestBetweenAnnotators(severity, instancesGroupedBySmells);
```
Luburic, line 20: For now, I would decoupled the Result class from our Domain Model. Currently, your (M)AnovaTest class returns a Result object, and these are domain services. We can have the application service wrap these results in the Result.Ok for now.

Luburic (changes requested):
Comments inline. They should be easy to solve, so feel free to merge when you are done. Before that make sure the tests are still working and fix them if they are not.

PR #60, opened 2021-07-03
Console DE app with configurable input

simsimkic:
The console app has 4 main functionalities:

1. Dataset creation

- enter the output folder path
- enter the projects (one or more) for which a dataset will be created (name of the project and project-commit url)
- a spreadsheet is created for each project

2. Dataset analysis

- enter the folder path with the annotated dataset
- enter the output folder path
2.1. Find insufficiently annotated instances
2.2. Find annotated instances with disagreeing annotations

3. Dataset exportation

- enter the projects (one or more) - local repo path and annotations folder path
- enter the path to the file with information about the annotators
- enter the output folder path
- the dataset is exported

4. Annotations consistency check

- enter the projects (one or more) - local repo path and annotations folder path
- enter the path to the file with information about the annotators
4.1. Consistency of a single annotator
4.2. Consistency between annotators
4.3. Significance of metrics for a single annotator
4.4. Significance of metrics between annotators
- enter the parameters such as annotator id or severity


DataSetExplorer/Program.cs
```diff
@@ -1,26 +1,254 @@
...
 {
     class Program
     {
+        private static IDataSetAnalyzer _dataSetAnalysisService = new DataSetAnalysisService();
+        
         static void Main(string[] args)
         {
-            //IDataSetCreator dataSetCreationService = new DataSetCreationService("C:\\main-repository\\", new GitCodeRepository());
-            //dataSetCreationService.CreateDataSetSpreadsheet("MyProject", "https://github.com/wieslawsoltes/Core2D/tree/50ea0849bef7e211500764ffcd1fece4a3cb224e");
+            string chosenOption;
```
Luburic, line 16: Extract into ConsoleUI and use Program for DI. I've discussed this a bit more in my other review.

Luburic (changes requested):

DataSetExplorer/ConsoleUI.cs
```diff
@@ -0,0 +1,293 @@
...
+using System.Collections.Specialized;
+using System.IO;
+using FluentResults;
+using System.Linq;
+
+namespace DataSetExplorer
+{
+    class ConsoleUI
+    {
+        private static IDataSetAnalyzerService _dataSetAnalysisService;
```
Luburic, line 14: static?

DataSetExplorer/ConsoleUI.cs
```diff
@@ -0,0 +1,293 @@
...
+                        break;
+                }
+            } while (!chosenOption.Equals("x"));
+        }
+
+        private static void ExportDataSet()
+        {
+            var projects = GetAnnotatedProjects();
+            var annotators = GetAnnotators();
+            IDataSetExporterService dataSetExportationService = new DataSetExportationService(new FullDataSetFactory(projects, annotators));
```
Luburic, line 138: Services should be initialized during Console UI construction. They should also be stateless, holding references only to those classes that do the work (because they are coordinators).

The FullDataSetFactory is probably also stateless, and if it is not, it should be instantiated for each request (by the service) so that the Service remains stateless. The projects and annotators should be supplied as input if they are inputs.

DataSetExplorer/ConsoleUI.cs
```diff
@@ -0,0 +1,293 @@
...
+        {
+            Console.WriteLine("\nChoose an option:");
+            Console.WriteLine("1. Check annotation consistency for a single annotator");
+            Console.WriteLine("2. Check annotation consistency between annotators");
+            Console.WriteLine("3. Check metrics significance for a single annotator");
+            Console.WriteLine("4. Check metrics significance between annotators");
+            Console.WriteLine("x. Exit\n");
+        }
+
+        private static void ChosenOptionErrorMessage(string chosenOption)
```
Luburic, line 288: This class is already quite large and there is room to grow. Does it make sense to extract a major submenu of this class into a separate class?

Luburic (changes requested):
We will discuss this tomorrow. It would be helpful if you can brush up on Clean Architecture https://www.youtube.com/watch?v=LcCRV-utBGU until then.

DataSetExplorer/ConsoleApp/ConsoleAppMainMenu.cs
```diff
@@ -0,0 +1,108 @@
+﻿using System;
+using System.Collections.Specialized;
+using FluentResults;
+
+namespace DataSetExplorer.ConsoleApp
+{
+    class ConsoleAppMainMenu
```
Luburic, line 7: Violation of DRY: MainMenu would suffice, as it is part of the ConsoleApp namespace.

DataSetExplorer/ConsoleApp/ConsoleIO.cs
```diff
@@ -0,0 +1,55 @@
...
+            Console.WriteLine("Option " + chosenOption + " not found. Choose again.\n");
+        }
+
+        internal static string GetAnswerOnQuestion(string question)
+        {
+            Console.Write(question);
+            return Console.ReadLine();
+        }
+
+        internal static ListDictionary GetAnnotatedProjects()
```
Luburic, line 22: This and the method below seem misplaced. Is there a better location to store them? If not leave them here, but add a comment that they should be moved somewhere, as they are not "generic enough" to belong to the broad ConsoleIO utility class.

Luburic (approved):
Looks a lot better now. A few minor comments to address and you can merge. Make sure the tests are passing first.

PR #62, opened 2021-07-18
Read database secrets from a file in order to utilize Docker Secrets

DanijelRadakovic:
This PR introduce minimal changes that needs to be done in order to utilize docker secrets. Also, platform setup is created via docker compose, so anyone can run and develop platform with ease.

Full automation process for platform setup requires configuration of Keycloak service. Currently, Keycloak needs to be configured manually when environment is created.


SETUP.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Building Images
+
+First step is to build images for Smart Tutor and Gateway services. Build configuration is set in `config/env.conf` file. 
+
+```shell
+docker-compose --env-file config/env.conf build
+```
+
+Default configuration creates `cleancadet/smart-tutor:lates` and `cleancadet/gateway:latest` images.
```
Luburic, line 23: `cleancadet/smart-tutor:lates` - are you missing a "t" at the end?

Luburic (approved):
One minor issue with the docs, but everything else looks ok.

Does this mean we are shutting down the operations branch (i.e., merging it into master)? I see the stacks folder is introduced in this PR so wanted to check if this was intended.

As a side note, a better branch name would be "feature/docker-secret-support" :)

SETUP.md
```diff
@@ -0,0 +1,171 @@
+# Setup
+
+This document provides steps that needs to be taken in order to successfully setup the infrastructure using Docker and Docker Compose.
+The infrastructure can be setup with or without Keycloak service (which provides SSO). In that matter there are two separate section 
+which describes infrastructure setup weather the Keycloak service is used.
+
+**Note**: For quick setup it is recommended to setup the infrastructure without
```
Luburic, line 7: I'll verify if these instructions are sound by going over this procedure and following it to the letter. I'll try to get around to this at the start of next week, but it might slip to the end of august because of my vacation.

SmartTutor/Startup.cs
```diff
@@ -126,7 +126,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             app.UseCors(x => x
                 .AllowAnyMethod()
                 .AllowAnyHeader()
-                .SetIsOriginAllowed(origin => true) // allow any origin
+                .SetIsOriginAllowed(_ => true) // allow any origin
```
Luburic, line 129: You should pull the changes from master that change this policy.

Luburic (approved):
As far as I can tell everything looks good. I'll need to check the instructions in more detail later, but we can easily update them once I get around to them.

It might be a good test to see how someone that is not familiar with the app setup fares against your instructions. @katarina-glorija we should talk about this PR today.

DanijelRadakovic:
@lukaDoric tested setup and it worked correctly.

PR #63, opened 2021-07-20
Migrate DE to Web application

rammba:
This PR contains _DbContext_ and minor changes in DE model.

### DIFF
--- a/DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
+++ b/DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
@@ -14,7 +14,6 @@ namespace DataSetExplorer.DataSetBuilder.Model
         public string ProjectLink { get; private set; }
         public SnippetType Type { get; private set; }
         public ISet<DataSetAnnotation> Annotations { get; private set; }
-        [NotMapped]
         public Dictionary<CaDETMetric, double> MetricFeatures { get; internal set; } // TODO: Expand and replace with the IFeature if a new feature type is introduced
 
         internal DataSetInstance(string codeSnippetId, string link, string projectLink, SnippetType type, Dictionary<CaDETMetric, double> metricFeatures)
--- a/DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
+++ b/DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
@@ -1,17 +1,20 @@
 ﻿using CodeModel.CaDETModel.CodeItems;
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations.Schema;
 using System.Linq;
 
 namespace DataSetExplorer.DataSetBuilder.Model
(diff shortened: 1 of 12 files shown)


DataSetExplorer/Database/DataSetExplorerContext.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public class DataSetExplorerContext : DbContext
+    {
+        public DataSetExplorerContext(DbContextOptions<DataSetExplorerContext> options) : base(options)
+        {
+        }
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+        {
+            modelBuilder.Entity<Annotator>().ToTable("Annotators");
```
Luburic, line 18: I don't think you need most of this code, if any.

image
Looking at the image above (from the SmartTutor DBContext), we define the fields of the DBContext so that we can use them to work with EFCore and the database. Simply by defining the fields we will already get a table when running using EF to generate our schema, as well as the appropriate foreign keys.

The reason why there is some configuration in our SmartTutor DBContext is mostly because of inheritance (e.g., the Images inherit the LearningObject class and should be placed in a separate table).

Try removing all the code from here, adding the appropriate fields and running add-migration on a new DB to see the results.

DataSetExplorer/Startup.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            app.UseAuthorization();
+
+            app.UseEndpoints(endpoints => { endpoints.MapControllers(); });
+        }
+
+        private string CreateConnectionStringFromEnvironment()
+        {
+            string server = Environment.GetEnvironmentVariable("DATABASE_HOST") ?? "localhost";
+            string port = Environment.GetEnvironmentVariable("DATABASE_PORT") ?? "5432";
```
Luburic, line 46: To avoid development frictions, we should maintain the same port and username and password, only a different schema name when compared to the SmartTutor db.

DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
```diff
@@ -1,17 +1,20 @@
...
-        public string ProjectLink { get; }
-        public SnippetType Type { get; }
-        public ISet<DataSetAnnotation> Annotations { get; }
+        public int Id { get; private set; }
+        public string CodeSnippetId { get; private set; }
+        public string Link { get; private set; }
+        public string ProjectLink { get; private set; }
+        public SnippetType Type { get; private set; }
+        public ISet<DataSetAnnotation> Annotations { get; private set; }
+        [NotMapped]
```
Luburic, line 17: Instead of polluting the domain model with persistence concerns, we should use the DBContext class to configure this.

image

Luburic (changes requested):
Comments inline. Once you address them make sure your tests for the Explorer are passing before requesting another review.

Luburic (approved):
Looks good.

PR #66, opened 2021-07-23
Create and analyze dataset use cases in DE web app

rammba:


DataSetExplorer/DataSetBuilder/DataSetInstanceService.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class DataSetInstanceService : IDataSetInstanceService
+    {
+        private readonly IDataSetInstanceRepository _dataSetInstanceRepository;
+
+        public DataSetInstanceService(IDataSetInstanceRepository dataSetInstanceRepository)
+        {
+            _dataSetInstanceRepository = dataSetInstanceRepository;
+        }
+
+        public Result<string> AddDataSetAnnotation(DataSetAnnotationDTO annotation, int annotatorId)
```
Luburic, line 22: Service classes should not work with DTOs but domain objects. Use AutoMapper https://docs.automapper.org/en/stable/Getting-started.html to transform the DTO into a domain object in the controller.

For examples, see e.g. https://github.com/Clean-CaDET/platform/blob/master/SmartTutor/Controllers/Progress/SubmissionController.cs#L34

Fix this across the board.

DataSetExplorer/DataSetBuilder/DataSetInstanceService.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            var instance = _dataSetInstanceRepository.GetDataSetInstance(annotation.DataSetInstanceId);
+            if (instance == default) throw new DataSetInstanceWithIdNotFound(annotation.DataSetInstanceId);
+            var annotator = _dataSetInstanceRepository.GetAnnotator(annotatorId);
+            if (annotator == default) throw new AnnotatorWithIdNotFound(annotatorId);
+            List<SmellHeuristic> smellHeuristics = new List<SmellHeuristic>();
+            foreach (var heuristic in annotation.ApplicableHeuristics) smellHeuristics.Add(new SmellHeuristic(heuristic.Description, true, heuristic.Reason));
+            DataSetAnnotation dataSetAnnotation = new DataSetAnnotation(annotation.CodeSmell, annotation.Severity, annotator, smellHeuristics);
+            instance.AddAnnotation(dataSetAnnotation);
+            _dataSetInstanceRepository.AddAnnotation(instance);
```
Luburic, line 32: AddAnnotation(instance) is not a good set of names. Update(instance) might be more suitable.

DataSetExplorer/DataSetBuilder/IDataSetAnalysisService.cs
```diff
@@ -1,10 +1,12 @@
...
+using System.Collections.Generic;
 
 namespace DataSetExplorer
 {
-    interface IDataSetAnalysisService
+    public interface IDataSetAnalysisService
     {
-        public Result<string> FindInstancesWithAllDisagreeingAnnotations(string dataSetPath, string outputPath);
-        public Result<string> FindInstancesRequiringAdditionalAnnotation(string dataSetPath, string outputPath);
+        public List<DataSetInstance> FindInstancesWithAllDisagreeingAnnotations(int dataSetId);
```
Luburic, line 9: Result

DataSetExplorer/DataSetBuilder/Model/Exceptions/AnnotatorWithIdNotFound.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace DataSetExplorer.DataSetBuilder.Model.Exceptions
+{
+    public class AnnotatorWithIdNotFound : Exception
```
Luburic, line 8: No need for these custom exceptions when using the Result pattern.

DataSetExplorer/Controllers/DataSetInstance/DataSetInstanceController.cs
```diff
@@ -0,0 +1,41 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace DataSetExplorer.Controllers.DataSetInstance
+{
+    [Route("api/dataset-instance/")]
+    [ApiController]
+    public class DataSetInstanceController : ControllerBase
```
Luburic, line 15: I'm not a fan of the "DataSetInstance" part of the name, since this controller deals with annotations.

AnnotationController might be better, and we can move the methods related to determining insufficient annotations here. Make these name changes across the board (e.g., for the service and repo).

Luburic (changes requested):
Several things to improve, comments inline.

DataSetExplorer/DataSetBuilder/DataSetAnalysisService.cs
```diff
@@ -30,16 +40,39 @@ public Result<string> FindInstancesRequiringAdditionalAnnotation(string dataSetP
...
+            return Result.Ok(instances);
+        }
+
         private DataSet LoadDataSet(string folder)
         {
             var importer = new ExcelImporter(folder);
             return importer.Import("Clean CaDET");
         }
+
+        private DataSet LoadDataSet(int dataSetId)
```
Luburic, line 72: No need for this method, as you've added 5 lines of code to replace _repository.GetDataSet(dataSetId) with GetDataSet(dataSetId). Furthermore, it is not clear how LoadDataSet(string) is different from LoadDataSet(int) from the function headers.

DataSetExplorer/Controllers/Dataset/DataSetController.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace DataSetExplorer.Controllers.Dataset
+{
+    [Route("api/dataset/")]
+    [ApiController]
+    public class DataSetController : ControllerBase
+    {
+        private readonly string _cloneGitPath = "../../ClonedProjects/";
```
Luburic, line 16: Should be read from configuration, e.g., https://medium.com/@dozieogbo/a-better-way-to-inject-appsettings-in-asp-net-core-96be36ffa22b

DataSetExplorer/Controllers/Dataset/DataSetController.cs
```diff
@@ -0,0 +1,52 @@
...
+            _dataSetCreationService = creationService;
+        }
+
+        [HttpPost]
+        public IActionResult CreateDataSet([FromBody] List<ProjectDTO> projects)
+        {
+            var dataSets = new List<DataSet>();
+            foreach (var project in projects)
+            {
+                var result = _dataSetCreationService.CreateDataSetInDatabase(_cloneGitPath, project.Name, project.Url);
```
Luburic, line 31: You'll need an async operation https://www.pluralsight.com/guides/using-task-run-async-await that you don't need to await - you just need to get the created Datasets with the status "Building" saved to the database, while the git clone is being conducted and the Code model is being compiled.

DataSetExplorer/DataSetBuilder/IDataSetCreationService.cs
```diff
@@ -1,11 +1,14 @@
...
 
 namespace DataSetExplorer
 {
     public interface IDataSetCreationService
     {
+        public Result<DataSet> CreateDataSetInDatabase(string basePath, string projectName, string projectAndCommitUrl);
         public Result<string> CreateDataSetSpreadsheet(string basePath, string projectName, string projectAndCommitUrl);
-        public Result<string> CreateDataSetSpreadsheet(string basePath, string projectName, string projectAndCommitUrl, NewSpreadSheetColumnModel columnModel);
+        public Result<string> CreateDataSetSpreadsheet(string basePath, string projectName, NewSpreadSheetColumnModel columnModel, DataSet dataSet);
+        public Result<DataSet> GetDataSetIfCreated(int id);
```
Luburic, line 12: "GetDataset" is sufficient - it will be created, but the question is what is it's status, and this is something the front-end will deal with.

DataSetExplorer/DataSetBuilder/Model/DataSet.cs
```diff
@@ -6,13 +6,15 @@ namespace DataSetExplorer.DataSetBuilder.Model
     public class DataSet
     {
         public int Id { get; private set; }
-        public string Url { get; }
-        internal readonly HashSet<DataSetInstance> _instances;
+        public string Url { get; private set; }
+        public HashSet<DataSetInstance> Instances { get; private set; }
+        public DataSetState State { get; set; }
```
Luburic, line 11: private set.

Luburic (changes requested):
A few comments to address, nothing major. When you complete it assign me as reviewer as well as @simsimkic

DataSetExplorer/Controllers/Dataset/DataSetController.cs
rammba: This is done in service: https://github.com/Clean-CaDET/platform/blob/feature/web-app-de/DataSetExplorer/DataSetBuilder/DataSetCreationService.cs#L26-L32
Is that a good solution?

DataSetExplorer/Startup.cs
```diff
@@ -14,11 +17,32 @@ namespace DataSetExplorer
...
+            services.AddAutoMapper(typeof(Startup));
+
             services.AddControllers();
+
             services.AddDbContext<DataSetExplorerContext>(opt =>
                 opt.UseNpgsql(CreateConnectionStringFromEnvironment()));
+
+            services.AddScoped<IDataSetCreationService, DataSetCreationService>();
+            services.AddScoped<ICodeRepository, GitCodeRepository>();
+            services.AddSingleton<IDataSetRepository, DataSetDatabaseRepository>();
```
simsimkic, line 40: Should be AddScoped instead of AddSingleton

rammba: I changed to _Transient_ now, because cannot update dataset after cloning repo with _Scoped_.

Luburic: This issue shouldn't be happening. @simsimkic can you pull the latest changes from this branch and see what is going on when you get a chance in the next few days?

Luburic (approved):
Looks good, ready for merge.

PR #67, opened 2021-07-26
Keycloak environment variables for smart-tutor. Keycloak setup instructions.

lukaDoric:


KEYCLOAK_SETUP.md
```diff
@@ -0,0 +1,56 @@
+# Setup Keycloak
+
+This document provides steps that need to be taken in order to successfully use Keycloak as IAM.
+
+Setup contains 9 steps:
+ - Adding a client.
```
Luburic, line 6: I would change this into an ordered list and use the appropriate numbers in each heading. Since "login as admin" is not a named step, I would integrate the related text into the introductory paragraphs.

Luburic (commented):
A minor detail to address, otherwise good job :)

DanijelRadakovic (approved):

PR #68, opened 2021-07-27
Keycloak role mapping.

lukaDoric:
This is the POC of how the user (in this case Learner) registration would look with IAM (in this case Keycloak) integrated in Smart-tutor. There are also some changes marked with TODO that will be done after merging #67 and #62 in master.


SmartTutor/SmartTutor.csproj
```diff
@@ -14,6 +14,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
```
Luburic, line 17: Why do we need this dependency? System.Text.Json should be quite enough, and I'm not a fan of introducing libraries when the base framework can provide all the functionality.

In general we should minimize additional code brought into our platform as that can cause issues with licensing. This is not the case for this particular package, but once again the base framework support should be enough.

SmartTutorTests/Integration/LearnerTests.cs
```diff
@@ -23,11 +24,12 @@ public void Successfully_logins()
         {
             using var scope = _factory.Services.CreateScope();
             var controller = new LearnerController(_factory.Services.GetRequiredService<IMapper>(),
-                scope.ServiceProvider.GetRequiredService<ILearnerService>());
+                scope.ServiceProvider.GetRequiredService<ILearnerService>(),
+                scope.ServiceProvider.GetRequiredService<IKeycloakService>());
```
Luburic, line 28: An example where we can inject the "NoAuthProvider" instead of the "KeycloakAuthProvider"

SmartTutor/Keycloak/KeycloakService.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using SmartTutor.LearnerModel.Learners;
+
+namespace SmartTutor.Keycloak
+{
+      public class KeycloakService : IKeycloakService
```
Luburic, line 12: KeycloakAuthProvider

SmartTutor/Keycloak/KeycloakService.cs
```diff
@@ -0,0 +1,97 @@
...
+            var keycloakUsers = JsonConvert.DeserializeObject<List<KeycloakLearner>>(await GetAllUsers());
+            foreach (var user in keycloakUsers.Where(s => s.Username.Equals(learner.StudentIndex)))
+            {
+                learner.IamId = user.Id;
+            }
+
+            return learner;
+        }
+    }
+}
```
Luburic, line 97: We should discuss this class on our next call.

Luburic (changes requested):
Some minor changes required, mostly names and packaging. Comments inline.

Luburic (approved):
Looks good. Once you add the docker secret support you can merge.

lukaDoric:
The GetSecret method was exported to the Util class so it could be available in other classes as well. @DanijelRadakovic @Luburic

DanijelRadakovic (approved):

Luburic (approved):

Luburic:
@lukaDoric Could you resolve conflict and merge?

PR #69, opened 2021-07-31
Make it stick recommender

milicasiriski:


SmartTutor/ContentModel/LearningObjects/LearningObject.cs
```diff
@@ -1,12 +1,16 @@
...
+﻿using System.Collections.Generic;
+using SmartTutor.ProgressModel.Progress;
+
+namespace SmartTutor.ContentModel.LearningObjects
 {
     public class LearningObject
     {
         public int Id { get; private set; }
         public int LearningObjectSummaryId { get; private set; }
+        public List<NodeProgress> NodeProgresses { get; set; }
```
Luburic, line 10: Can we remove this association?

There is no use case where we will want to access all the node progresses that contain a particular LO. Furthermore, the content model should be minimally aware of the progress model, and this adds a dependency that we don't need.

You can setup at the EF Code configuration level (SmartTutorContext) the many-to-many relationship.

SmartTutor/ContentModel/Lectures/KnowledgeNode.cs
```diff
@@ -9,9 +9,12 @@ public class KnowledgeNode
         public KnowledgeNodeType Type { get; private set; }
         public int LectureId { get; private set; }
         public List<LearningObjectSummary> LearningObjectSummaries { get; private set; }
+        public List<KnowledgeNode> Prerequisites;
```
Luburic, line 12: Are we missing {get; private set;} here? Also, can we add private to the setter of PrerequisiteOf?

SmartTutor/InstructorModel/Instructors/MakeItStickRecommender.cs
```diff
@@ -0,0 +1,62 @@
...
+using SmartTutor.ContentModel.LearningObjects;
+using SmartTutor.ContentModel.LearningObjects.Repository;
+using SmartTutor.ContentModel.Lectures;
+using SmartTutor.ContentModel.Lectures.Repository;
+
+namespace SmartTutor.InstructorModel.Instructors
+{
+    public class MakeItStickRecommender : IInstructor
+    {
+        private readonly VARKRecommender _varkRecommender;
```
Luburic, line 13: Instead of VARKRecommender why not IInstructor?

SmartTutor/InstructorModel/Instructors/MakeItStickRecommender.cs
```diff
@@ -0,0 +1,62 @@
...
+            return _varkRecommender.GatherDefaultLearningObjects(learningObjectSummaries);
+        }
+
+        private void AddPrerequisiteRevisionQuestions(int id, List<LearningObject> result)
+        {
+            Random random = new Random();
+            KnowledgeNode knowledgeNode = _lectureRepository.GetKnowledgeNode(id);
+            foreach (var prerequisiteNode in knowledgeNode.Prerequisites ?? new List<KnowledgeNode>())
+            {
+                int index = random.Next(0, prerequisiteNode.LearningObjectSummaries.Count);
```
Luburic, line 46: I don't understand the random aspect here?

SmartTutor/ContentModel/Lectures/KnowledgeNode.cs
```diff
@@ -9,9 +9,12 @@ public class KnowledgeNode
...
+        //TODO: Remove if EF6 supports unidirectional many-to-many
+        public List<KnowledgeNode> PrerequisiteOf { get; set; }
```
Luburic, line 15: We talked about adding a "recap learning object summary" - essentially each knowledge node could mark a single learning object summary that would serve as the "recap" for the knowledge node. This recap LOSummary could then be retrieved by the recommender.

SmartTutor/Startup.cs
```diff
@@ -66,7 +66,8 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<IWorkspaceCreator, NoWorkspaceCreator>();
             services.AddScoped<ILearnerRepository, LearnerDatabaseRepository>();
 
-            services.AddScoped<IInstructor, VARKRecommender>();
+            services.AddScoped<VARKRecommender, VARKRecommender>();
```
Luburic, line 69: This article should help with the composite IInstructor https://andrewlock.net/adding-decorated-classes-to-the-asp.net-core-di-container-using-scrutor/ - make note of it, as it will serve as an interesting segment in your BSc.

Luburic (changes requested):
A few things to change, comments inline.

SmartTutor/ContentModel/LearningObjects/LearningObject.cs
milicasiriski: Yes, I agree. I've used this to fix the issue where the Learning Object was being connected to only one Node Progress, which was causing issues when a knowledge node is loaded multiple times (and also wasn't consistent with the intended model). I'll configure the right relationship in the the SmartTutorContext 😄

SmartTutor/Database/Scripts/all.sql
```diff
@@ -27,6 +28,8 @@ DELETE FROM public."ArrangeTaskContainers";
 DELETE FROM public."ArrangeTasks";
 DELETE FROM public."LearningObjects";
 DELETE FROM public."LearningObjectSummaries";
+-- TODO: Learning objects reference Node Progress so the following line had to be moved here
```
Luburic, line 31: Are these TODOs done?

SmartTutor/Startup.cs
```diff
@@ -81,7 +82,9 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<IWorkspaceCreator, NoWorkspaceCreator>();
             services.AddScoped<ILearnerRepository, LearnerDatabaseRepository>();
 
-            services.AddScoped<IInstructor, VARKRecommender>();
+            services.AddScoped<IPrerequisiteSelectionStrategy, RandomizedPrerequisiteSelectionStrategy>();
+            services.AddScoped<IInstructor, VARKInstructor>();
```
Luburic, line 86: Could you explain how lines 86 and 87 work together?

This is something we should have in the BSc as well, to explain the whole configuration setup.

Luburic (changes requested):
A few comments inline plus three additional questions:
- Will we have some tests to check whether the MakeItStickInstructor works?
- Did we resolve the reported issue with array index out of bounds?
- We still need real questions for the recap LOs, right?

SmartTutor/Startup.cs
milicasiriski: Yes, I've read the article and am going to include it in my thesis :smile:

SmartTutor/Database/Scripts/all.sql
milicasiriski: Yes, I will go through all of the changes again to make sure everything is done

Luburic:
Closing this PR as we have migrated the Tutor to a separate solution.

PR #71, opened 2021-08-05
Add DataSetProject

simsimkic:


DataSetExplorer/DataSetBuilder/Model/DataSetProject.cs
```diff
@@ -0,0 +1,46 @@
+using System.Collections.Generic;
+using System.Linq;
+
+namespace DataSetExplorer.DataSetBuilder.Model
+{
+    public class DataSetProject
+    {
+        internal readonly string _name;
```
Luburic, line 8: We use the "_" prefix to signal private fields, and these are internal.

Since we will need to map them to DTOs, we should stick with properties with exposed getters and private setters.

Luburic (changes requested):
Several comments inline. Also ensure you are able to create a db-schema and that the exposed endpoints are working (use postman if not the Angular UI) to check.

Luburic (approved):

PR #73, opened 2021-08-13
Parse CORS origins from .txt file.

lukaDoric:


SmartTutor/Startup.cs
```diff
@@ -171,5 +171,17 @@ private static string CreateConnectionStringFromEnvironment()
             return
                 $"Server={server};Port={port};Database={database};User ID={user};Password={password};Integrated Security={integratedSecurity};Pooling={pooling};";
         }
+
+        private static string[] ParseCorsOrigins()
+        {
+            string[] corsOrigins = { "http://localhost:4200" };
```
DanijelRadakovic, line 177: I get the idea to allow front-end web app access by default, but fixing this in code is not a good practice and not secure either. My proposal is to remove this line of code and put URL in cors.txt file. This should be removed and put in cors.txt file. Setup.md file should be changed also by adding some instructions how to setup CORS.

DanijelRadakovic: I am also suggesting creation of configuration file or files for Smart Tutor service in order to remove all magical URLs, usernames, ports etc. from source code and reduce the number of environment variables. In that way, separation between source code and its configuration would be much clearer.

DanijelRadakovic (approved):

SmartTutor/Startup.cs
Luburic: Yeah, we should move this to configuration definitely.

Luburic (approved):
Comment inline. We should move the CORS configuration

PR #74, opened 2021-08-22
Feature: Identify cohesive parts

balsa-sarenac:
Given a class, try to split it by removing method-field interactions. Returns a list of potential refactoring options, each containing:
- a list of interactions to remove to perform refactoring
- two classes that are the result of the refactoring and are also highly cohesive


CodeModel/CodeParsers/CSharp/Exceptions/CohesionAnalyzerException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace CodeModel.CodeParsers.CSharp.Exceptions
+{
+    public class CohesionAnalyzerException : Exception
```
Luburic, line 5: This name is generic and doesn't reveal what the problem is.

CodeModel.Tests/Unit/CodeParser/CohesionAnalyzerTests.cs
```diff
@@ -0,0 +1,48 @@
...
+using CodeModel.CodeParsers.CSharp.ProjectAnalyzer;
+
+namespace CodeModel.Tests.Unit.CodeParser
+{
+    public class CohesionAnalyzerTests
+    {
+        private static readonly CodeFactory TestDataFactory = new();
+
+        [Fact]
+        public void Test_Cohesion_Analyzer()
```
Luburic, line 15: Meaningful test name - when the test fail I should ideally know exactly which function is misbehaving and in which case.

CodeModel.Tests/Unit/CodeParser/CohesionAnalyzerTests.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            var factory = new CodeModelFactory();
+            var classes =
+                factory.CreateProject(TestDataFactory.GetCohsionClasses()).Classes;
+
+            var analyzer = new CohesionAnalyzer();
```
Luburic, line 21: Theory.

CodeModel/CodeParsers/CSharp/ProjectAnalyzer/ClassInteractions.cs
```diff
@@ -0,0 +1,104 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using CodeModel.CodeParsers.CSharp.Exceptions;
+
+namespace CodeModel.CodeParsers.CSharp.ProjectAnalyzer
+{
+    public class ClassInteractions
```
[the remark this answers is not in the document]
Luburic: Yes. Always aim for simplicity.

CodeModel/CodeParsers/CSharp/ProjectAnalyzer/ClassInteractions.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        private int[,] MethodFieldAccessMapping { get; }
```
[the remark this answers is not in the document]
Luburic: Or MethodFieldAccesses if it contains information on which methods access which fields? We can go with Interactions if we use the term consistently and define what we mean by interaction. Access is maybe a more specific and established term though.

CodeModel/CodeParsers/CSharp/ProjectAnalyzer/Edge.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CodeModel.CodeParsers.CSharp.ProjectAnalyzer
+{
+    public class Edge
```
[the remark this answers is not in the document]
Luburic: Maybe, but do we have some literature to draw on to justify this term?

Luburic (changes requested):
Comments inline.

CodeModel.Tests/DataFactories/CodeFactory.cs
```diff
@@ -66,6 +66,7 @@ internal bool IsAvailable(DateRange timeSpan)
             }"
             };
         }
```
balsa-sarenac, line 68: @Luburic I've formatted this file by mistake. Should this be reverted, or it's fine to say?

CodeModel/CodeParsers/CSharp/ProjectAnalyzer/Access.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+namespace CodeModel.CodeParsers.CSharp.ProjectAnalyzer
```
balsa-sarenac, line 3: Currently, the cohesion analyzer is inside the `ProjectAnalyzer` directory. I was thinking of renaming the directory to `ClassCohesionAnalyzer`. What do you think?

CodeModel/CodeParsers/CSharp/ProjectAnalyzer/CohesiveParts.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.Generic;
+
+namespace CodeModel.CodeParsers.CSharp.ProjectAnalyzer
+{
+    public class CohesiveParts
+    {
+        public List<HashSet<Access>> Parts { get; }
```
balsa-sarenac, line 7: Refactoring proposition: create a wrapper class around `HashSet<Access>`. Many methods have this collection as a parameter. Also by doing it I would avoid using `List<HashSet<Access>>`, and have a more meaningful name.
I am not sure about the class name. I was thinking about `Part` since it represents a part of a class, could be whole class, could be few method-field interactions. Maybe `ClassPart` since `Part` is too generic?

CodeModel/CodeParsers/CSharp/ClassCohesionAnalyzer/CohesiveParts.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.Generic;
+
+namespace CodeModel.CodeParsers.CSharp.ClassCohesionAnalyzer
+{
+    public class CohesiveParts
+    {
+        public List<HashSet<Access>> Parts { get; }
```
balsa-sarenac, line 7: Refactoring proposition: create a wrapper class around `HashSet<Access>`. Many methods have this collection as a parameter. Also by doing it I would avoid using `List<HashSet<Access>>`, and have a more meaningful name.
I am not sure about the class name. I was thinking about `Part` since it represents a part of a class, could be whole class, could be few method-field interactions. Maybe `ClassPart` since `Part` is too generic?

CodeModel.Tests/DataFactories/CodeFactory.cs
Luburic: It's fine.

CodeModel/CodeParsers/CSharp/ClassCohesionAnalyzer/CohesionAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+                if (parts is not { Count: 2 }) continue;
+                var cohesiveParts = new CohesiveParts(accesses, parts);
+                possibleParts.RemoveAll(s => s.AccessesToCut.IsProperSupersetOf(cohesiveParts.AccessesToCut));
+                possibleParts.Add(cohesiveParts);
+            }
+
+            return possibleParts;
+        }
+
+        private IEnumerable<HashSet<Access>> GetAccessesThatCanBeCut(HashSet<Access> allAccesses,
```
Luburic, line 38: I find it a bit odd that the methods GetAccessesThatCanBeCut and GetAccessesThatCannotBeCut are in separate classes.

CodeModel/CodeParsers/CSharp/ClassCohesionAnalyzer/CohesionAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+            IEnumerable<HashSet<Access>> accessesThatCannotBeCut)
+        {
+            var data = allAccesses.ToArray();
+            if (data.Length == 0) return new List<HashSet<Access>>();
+
+            var allAccessesCombinations = Enumerable
```
Luburic, line 44: This is a heavy expression that is difficult to understand.

CodeModel/CodeParsers/CSharp/ClassCohesionAnalyzer/CohesiveParts.cs
Luburic: This could probably host the GetAccessesThatCanBeCut and similar methods?

Luburic (commented):
It's a lot more clear than before.

One other thing to consider - how easy is it to change the algorithm for the metric that we apply at the second level (currently LCOM)? From what I can tell, this is hard coded and not really interchangable. Maybe the _strategy_ pattern can be useful for this segment?

CodeModel/CodeParsers/CSharp/ClassCohesionAnalyzer/CohesiveParts.cs
balsa-sarenac: Yes, most of the methods that require or work with `HashSet<Access>`.
One more thing, when I do refactor this, class `ClassInteractions` will only have a responsibility to create the initial `HashSet<Access>`, and I'm thinking of turning it into a static factory method. Thoughts?

Luburic: The static factory method can be part of the ClassPart class :)

CodeModel/CodeParsers/CSharp/ClassCohesionAnalyzer/CohesionAnalyzer.cs
balsa-sarenac: I have created my own implementation of this method that is slower than this one. That's the reason I left it out. Implementation was a loop iterating through a number of elements that we want in a group (0, n/2 - 1). Inside the loop is a call to a recursive function that finds all combinations for a given number of elements. It should be faster with caching but I am not sure if it'll be as fast as this.
Any suggestions?
Should I put this in a function and write some inline comments explaining some details?

Luburic: A comment might be sufficient. It's ok to have complex code when there is a significant performance gain, and performance is an issue with this algorithm.

Luburic:
Closing for now. We might come back to this research direction some day.

PR #77, opened 2021-09-01
Code metrics

simsimkic:
- [x] **WOC - Weight Of a Class**
From book '_Object-oriented metrics in practice_':
image

- [x] **AID - Access of Import Data** (Counts attributes and methods. Differs from ATFD which counts attributes and get/set methods.)
- [x] **NOPA - Number Of Public Attributes**
From book '_Object-oriented metrics in practice_':
image

- [x] **NOPP - Number Of Public Properties** (A simpler version of the metric **NOAM - Number Of Accessor Methods** (getter and setter), due to the current CaDET model implementation on the Platform)
Our implementation: Ignore the exact number of public getters and setters, and count only public properties.
NOAM from book '_Object-oriented metrics in practice_':
image

- [x] **WMCNAMM - Weighted Methods Count of Not Accessor or Mutator Methods** (The sum of complexity of the methods that are defined in the class, and are not accessor or mutator methods. We compute the complexity with the Cyclomatic Complexity metric (CYCLO))
- [x] **BUR - Base class Usage Ratio** (The number of inheritance-specific members used by the measured class divided by the total number of inheritance-specific members from the base class.)
- [x] **BOvR - Base class Overriding Ratio** (The number of methods of the measured class that override methods from base class, divided by the total number of methods in the class.)

### DIFF
--- a/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
+++ b/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
@@ -149,6 +149,12 @@
         /// </summary>
         WMCNAMM,
 
+        /// <summary>
+        /// BUR: Base class Usage Ratio, counts the number of inheritance-specific members used by the measured class, 
+        /// divided by the total number of inheritance-specific members from the base class
+        /// </summary>
+        BUR,
+
         #endregion
 
         #region MemberMetrics
--- a/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
+++ b/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
@@ -139,6 +139,11 @@
         /// </summary>
         NOPA,
 
+        /// <summary>
+        /// NOAM: Number Of Public Properties
+        /// </summary>
+        NOPP,
+
         #endregion
 
         #region MemberMetrics
--- a/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
+++ b/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
@@ -155,6 +155,12 @@
         /// </summary>
         BUR,
 
+        /// <summary>
+        /// BOvR: Base class Overriding Ratio, counts the number of methods of the measured class that override methods from the base class, 
+        /// divided by the total number of methods in the class
+        /// </summary>
+        BOvR,
+
         #endregion
 
         #region MemberMetrics
--- a/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
+++ b/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
@@ -134,9 +134,14 @@
         /// </summary>
         WOC,
 
-    #endregion
+        /// <summary>
+        /// NOPA: Number Of Public Attributes
+        /// </summary>
+        NOPA,
 
-    #region MemberMetrics
+        #endregion
+
+        #region MemberMetrics
         /// <summary>
         /// CYCLO - Cyclomatic complexity
         ///
--- a/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
+++ b/CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
@@ -129,9 +129,14 @@
         /// </summary>
         NIC,
 
-        #endregion
+        /// <summary>
+        /// WOC: Weight Of a Class, counts the number of “functional” public methods divided by the total number of public members
+        /// </summary>
+        WOC,
 
-        #region MemberMetrics
+    #endregion
+
+    #region MemberMetrics
         /// <summary>
         /// CYCLO - Cyclomatic complexity
         ///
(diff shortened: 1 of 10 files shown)


CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -37,10 +37,127 @@ internal class CaDETClassMetricCalculator
...
+            var staticFields = (from field in publicFields
+                             from modifier in field.Modifiers
+                             where modifier.Value.Equals(CaDETModifierValue.Static)
+                             select field).ToList();
+            return publicFields.Where(f => !staticFields.Contains(f)).Count();
+        }
+
+        private static double CountWeightOfClass(CaDETClass parsedClass)
+        {
+            return (double)CountFunctionalPublicMethods(parsedClass.Members) / (CountPublicProperties(parsedClass.Members) + GetPublicFields(parsedClass.Fields).Count());
```
Luburic, line 118: You should try to keep the ordering of the private functions to follow the expression. For this method, I would put CountFunctionalPublicMethods, then CountPublicProperties, and then GetPublicFields. It's easier to read that way.

CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -37,10 +37,127 @@ internal class CaDETClassMetricCalculator
...
+        {
+            var methods = GetMethods(members);
+            var publicMethods = FindMembersWithModifier(methods, CaDETModifierValue.Public);
+            var abstractMethods = FindMembersWithModifier(publicMethods, CaDETModifierValue.Abstract);
+            return publicMethods.Where(m => !abstractMethods.Contains(m)).Count();
+        }
+
+        private static IEnumerable<CaDETMember> FindMembersWithModifier(IEnumerable<CaDETMember> members, CaDETModifierValue modifierValue)
+        {
+            return (from member in members
```
Luburic, line 155: Don't use this syntax. While there is nothing wrong with it, it is a coding style that we do not follow in the rest of the project and we shouldn't mix styles.

If you need a utility function in CaDETMember or CaDETModifier add it.

CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -37,10 +37,127 @@ internal class CaDETClassMetricCalculator
...
+            var properties = members.Where(m => m.Type.Equals(CaDETMemberType.Property));
+            var publicProperties = FindMembersWithModifier(properties, CaDETModifierValue.Public);
+            return publicProperties.Count();
+        }
+
+        // Functional public methods do not include get/set, constructor and abstract methods. (Object-oriented metrics in practice)
+        private static int CountFunctionalPublicMethods(List<CaDETMember> members)
+        {
+            var methods = GetMethods(members);
+            var publicMethods = FindMembersWithModifier(methods, CaDETModifierValue.Public);
```
Luburic, line 148: Maybe not for this PR, but something to keep in mind - maybe the GetMethods method of CaDETClass can have optional parameters - CaDETModifierValue withModifier, CaDETModifierValue withoutModifier (or maybe even lists of these).

We will see how new metrics utilize modifiers and refactor accordingly.

CodeModel/CodeParsers/CSharp/CaDETClassMetricCalculator.cs
```diff
@@ -37,10 +37,127 @@ internal class CaDETClassMetricCalculator
...
+                [CaDETMetric.WOC] = CountWeightOfClass(parsedClass),
+                [CaDETMetric.NOPA] = CountPublicAttributes(parsedClass),
+                [CaDETMetric.NOPP] = CountPublicProperties(parsedClass.Members),
+                [CaDETMetric.WMCNAMM] = GetWMCOfNotAccessorOrMuttatorMethods(parsedClass),
+                [CaDETMetric.BUR] = GetBaseClassUsageRatio(parsedClass),
+                [CaDETMetric.BOvR] = GetBaseClassOverridingRatio(parsedClass),
             };
         }
 
+        private static double GetBaseClassOverridingRatio(CaDETClass parsedClass)
```
Luburic, line 50: Since this is the last calculated metric I would expect to find the method at the bottom of the whole class. Same goes for the rest of the methods, ordering is important.

Luburic (changes requested):
Several comments inline. Many are future work, but a few should be addressed now.

CodeModel/CaDETModel/CodeItems/CaDETMetrics.cs
```diff
@@ -229,7 +261,12 @@ public enum CaDETMetric
         /// <summary>
         /// NOUW: Number of unique words
         /// </summary>
-        NOUW
+        NOUW,
+
+        /// <summary>
+        /// AID: Access to import data, counts attributes and methods (differs from ATFD which counts attributes and get/set methods)
+        /// </summary>
+        AID,
```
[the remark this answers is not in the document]
simsimkic: DE saves metrics by the name.

Luburic (approved):
Looks good. Feel free to merge if the tests are passing.

PR #81, opened 2021-09-22
Add CandidateInstance and smells automatization rules

simsimkic:
CandidateInstance represents a list of instances related to a particular type of code smell. 
Project contains list of candidate instances.


DataSetExplorer/DataSetBuilder/Model/CandidateDataSetInstance.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Collections.Generic;
+
+namespace DataSetExplorer.DataSetBuilder.Model
+{
+    public class CandidateDataSetInstance
```
Luburic, line 5: We need a better name. First, I would like to remove the "DataSet" keywords from most of the model. We need it for the high-level DataSet, and I think it makes sense to have the "DataSetProject". We should remove it from this and the pure Instance class.

Now, regarding this class name, a better alternative might be `SmellCandidateInstances` or `CandidateSmellInstances` or `SmellInstanceGroup`. Maybe the first one, as Candidate is a notable name in the domain of annotation.

As a side note, this will be an interesting point for your paper, where the expanded conceptual model is described and the reasoning behind it. Maybe create a google doc or Clickup card where you will keep track of all the things you'll discuss in your paper.

DataSetExplorer/DataSetBuilder/Model/DataSet.cs
```diff
@@ -8,11 +7,13 @@ public class DataSet
         public int Id { get; private set; }
         public string Name { get; private set; }
         public HashSet<DataSetProject> Projects { get; private set; }
-        
-        public DataSet(string name)
+        public List<CodeSmell> CodeSmells { get; private set; }
```
Luburic, line 10: Maybe a more descriptive name? What does this list of CodeSmells mean for the DataSet? `SupportedSmells`? `ExaminedSmells`?

DataSetExplorer/DataSetBuilder/Model/DataSetInstance.cs
```diff
@@ -23,14 +23,24 @@ internal DataSetInstance(string codeSnippetId, string link, string projectLink,
...
             Validate();
         }
 
         private DataSetInstance()
         {
         }
 
+        private void SetMetricFeatures(Dictionary<CaDETMetric, double> metricFeatures)
+        {
+            if (metricFeatures == null) return;
```
Luburic, line 36: You can integrate this logic into the setter (https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-properties).

DataSetExplorer/DataSetBuilder/Model/DataSetProject.cs
```diff
@@ -9,42 +8,60 @@ public class DataSetProject
...
             State = DataSetProjectState.Processing;
         }
 
         public DataSetProject(string name) : this(name, null) { }
 
-        internal void AddInstances(List<DataSetInstance> instances)
+        internal void AddCandidateInstance(CandidateDataSetInstance newCandidate)
         {
-            foreach (var instance in instances)
+            var i = CandidateInstances.FindIndex(c => c.CodeSmell.Name.Equals(newCandidate.CodeSmell.Name));
```
Luburic, line 26: You can override Equals (and don't forget hashcode) for CandidateInstances to check if the code smell name is the same. Then you can use a HashSet to automatically perform this check (similar to the earlier code).

We can discuss this if you are unsure.

DataSetExplorer/DataSetBuilder/Model/DataSetProject.cs
```diff
@@ -56,5 +73,10 @@ public void Failed()
         {
             if (State == DataSetProjectState.Processing) State = DataSetProjectState.Failed;
         }
+
+        internal void SetCandidateInstances(List<CandidateDataSetInstance> candidateInstances)
```
Luburic, line 77: No need for these methods when you have properties.

DataSetExplorerTests/Unit/ImporterTests.cs
```diff
@@ -13,30 +13,37 @@ public class ImporterTests
         [Fact]
         public void Imports_data_set_instances_and_annotations()
         {
-            var dataSet = new DataSet("Test");
+            var codeSmells = new List<CodeSmell>
```
Luburic, line 16: How good is our test coverage for the dataset explorer functionality? Something to keep in mind.

Luburic (changes requested):
Comments inline.

DataSetExplorer/DataSetSerializer/ViewModel/ColumnHeuristicsModel.cs
```diff
@@ -12,8 +12,11 @@ internal ColumnHeuristicsModel()
         {
             _heuristics = new Dictionary<CodeSmell, List<string>>();
             //TODO: Load from DB or config.
```
Luburic, line 14: Since you've already introduced this information in the appsettings.Development.json, would it be better to load the information from there?

If I'm not mistaken, we will refactor both this and the appsettings with this story https://app.clickup.com/t/u9nd8j so you can leave this for now.

DataSetExplorer/DataSetBuilder/Model/Instance.cs
```diff
@@ -5,43 +5,68 @@
...
     {
         public int Id { get; private set; }
         public string CodeSnippetId { get; private set; }
         public string Link { get; private set; }
         public string ProjectLink { get; private set; }
         public SnippetType Type { get; private set; }
-        public ISet<DataSetAnnotation> Annotations { get; private set; }
-        public Dictionary<CaDETMetric, double> MetricFeatures { get; internal set; } // TODO: Expand and replace with the IFeature if a new feature type is introduced
+        public ISet<Annotation> Annotations { get; private set; }
+       /* public Dictionary<CaDETMetric, double> MetricFeatures
```
Luburic, line 16: Should this be removed?

Luburic (changes requested):
Several comments inline to address.

DataSetExplorer/DataSetBuilder/InstanceFilter.cs
```diff
@@ -0,0 +1,50 @@
...
+            var filteredInstances = new List<Instance>();
+            var metricsForSmell = MetricThresholds.FindAll(m => m.CodeSmell.Equals(codeSmell.Name));
+
+            foreach (var i in instances)
+            {
+                if (!ValidSnippetTypeForSmell(i, codeSmell)) continue;
+                var fulfilledConditions = new List<bool>();
+                foreach (var thresholds in metricsForSmell)
+                {
+                    fulfilledConditions.Add(InstanceFulfillesConditions(i, thresholds));
```
Luburic, line 28: Can this be simplified to:

    if (!ValidSnippetTypeForSmell(i, codeSmell)) continue;
    if (InstancePassesMetricThresholds(i, metricsForSmell)) filteredInstances.Add(i);


    InstancePassesMetricThresholds can have code similar to "return metricsForSmell.All(metricThreshold => InstanceFulfillesConditions(i, metricThreshold));"

Luburic (changes requested):
A few minor comments to address.

PR #83, opened 2021-10-07
Export members from annotated classes

simsimkic:


DataSetExplorer/DataSetBuilder/DataSetAnalysisService.cs
```diff
@@ -72,5 +75,46 @@ private DataSetProject LoadDataSetProject(string folder, string projectName)
...
+            CodeModelFactory factory = new CodeModelFactory();
+
+            try
+            {
+                var classesGroupedBySeverity = new Dictionary<int, List<CaDETClass>>();
+                var annotatedInstances = LoadAnnotatedInstances(datasetPath);
+
+                foreach (var projectUrl in projects.Keys)
+                {
+                    CaDETProject cadetProject = factory.CreateProjectWithCodeFileLinks(projects[projectUrl]);
```
Luburic, line 96: Is there room for extracting a method here to reduce the overall complexity? I would avoid creating a new method that takes 3-4 parameters or has 2-3 LOC, but if we can get away with extracting most of this foreach body with 1 or 2 parameters that would be great.

PR #87, opened 2021-11-15
Export draft dataset

simsimkic:


DataSetExplorer/Controllers/Dataset/DataSetController.cs
```diff
@@ -23,6 +24,15 @@ public DataSetController(IMapper mapper, IDataSetCreationService creationService
             _gitClonePath = configuration.GetValue<string>("Workspace:GitClonePath");
         }
 
+        [HttpPost]
+        [Route("export")]
+        public IActionResult ExportDataSet([FromBody] DraftDataSetExportDTO dataSetDTO)
+        {
+            var dataSet = _dataSetCreationService.GetDataSet(dataSetDTO.Id).Value;
+            var exportPath = new DraftDataSetExporter(dataSetDTO.ExportPath).Export(dataSetDTO.AnnotatorId, dataSet);
+            return Ok(new FluentResults.Result().WithSuccess("Successfully exported to " + exportPath));
```
Luburic, line 33: This will "work on my machine" but is not a good solution if you have a web application that is hosted on a remote server whose file system we cannot access. We should get a link to the file to activate the download process on the front-end app.

Luburic (commented):
This works for our case, but is not a good solution for a production system. See if you want to address this issue now.

DataSetExplorer/Controllers/Dataset/DataSetController.cs
simsimkic: This comment will be addresesd as part of this task https://app.clickup.com/t/1tk1dy3.

PR #89, opened 2021-11-16
Delete&update datasets and projects

simsimkic:


DataSetExplorer/DataSetBuilder/Model/Repository/DataSetDatabaseRepository.cs
```diff
@@ -59,5 +59,50 @@ public void Update(DataSet dataSet)
...
+        {
+            var deletedProject = _dbContext.DataSetProjects.Remove(FindProjectForDeletion(id)).Entity;
+            _dbContext.SaveChanges();
+            return deletedProject;
+        }
+
+        private DataSetProject FindProjectForDeletion(int id)
+        {
+            return _dbContext.DataSetProjects
+                .Include(p => p.CandidateInstances).ThenInclude(c => c.Instances).ThenInclude(i => i.Annotations).ThenInclude(a => a.ApplicableHeuristics)
```
Luburic, line 96: Not sure how EF handles this, but do you need all these includes if you configured cascade deletion below?

Luburic (approved):
Looks good, with one thing to consider.

DataSetExplorer/DataSetBuilder/Model/Repository/DataSetDatabaseRepository.cs
simsimkic: You are right, i don't need this. I added some things that were missing in the cascade delete configuration.

PR #91, opened 2021-12-01
Fix NOUW metric calculation

simsimkic:
Include constructors (the previous implementation was only for methods).
Ignore case when counting unique words.


Luburic (approved):
Looks good.

PR #94, opened 2021-12-09
Refactor DataSetBuilder namespace

simsimkic:


Luburic (commented):
Consider the following questions/comments and contemplate if it makes sense to change the code accordingly:

1. Would the project improve if you added a few folders at the top level folder (DataSetExplorer), that can group the code according to its function. For example, the folders "Core" (containing business logic), "Infrastructure" (containing technical details related to the Database, RepositoryAdapters), and "UI" (Controllers and ConsoleApp) might help decluter the starting folder. Examine https://github.com/Clean-CaDET/tutor/tree/master/src for inspiration.
2. You have the Annotations namespace, and yet the Annotation model is located in the DataSets namespace. Would it be better to create 3 modules in your Core domain, that correspond to the 3 segments of the conceptual model we used in our paper (i.e., 1 - Dataset, Dataset Project, Dataset Instance; 2 - Schema - Code smell, Code smell heuristic; 3 - Annotation, Annotator. Where would we fit the Metric Thresdholds, and Smell Filter? If not 3 packages, maybe 2 or 4?
3. The repository namespace should probably be at the module-level (i.e., inside the DataSets folder instead of the Model folder).

simsimkic:
1. Done
2. I've skipped the Schema folder for now. I'll work on this in the annotation schema PR.
3. Done

Luburic (approved):
Looks good.

PR #96, opened 2021-12-27
Related instances for annotated instance

simsimkic:


DataSetExplorer/Core/DataSets/CaDETToDataSetProjectBuilder.cs
```diff
@@ -26,6 +26,7 @@ internal class CaDETToDataSetProjectBuilder
         private CaDETMemberType[] _acceptedMemberTypes = {CaDETMemberType.Constructor, CaDETMemberType.Method};
         private readonly List<CodeSmell> _codeSmells;
         private readonly InstanceFilter _instanceFilter;
+        private Dictionary<CaDETClass, Dictionary<CaDETClass, int>> _classReferences = new Dictionary<CaDETClass, Dictionary<CaDETClass, int>>();
```
Luburic, line 29: Maybe Dictionary<CaDETClass, List<CoupledClassStrength>> where CoupledClassStrength contains CaDETClass and int?

The _classReference is a bit generic, but I'm not sure what a more suitable name looks like (maybe _classCouplings?)

DataSetExplorer/Core/DataSets/CaDETToDataSetProjectBuilder.cs
```diff
@@ -110,9 +111,85 @@ private int DetermineNumberOfInstances(int totalNumber)
...
+                {
+                    AddReferenceToMap(reference, instance);
+                }
+            }
+        }
+
+        private static Dictionary<CaDETClass, int> GetReferencedInstances(CaDETClass c)
+        {
+            var referencedInstances = new List<CaDETClass>();
+            referencedInstances.AddRange(c.GetFieldLinkedTypes());
```
Luburic, line 135: My issue with understanding this code is that you are using "reference" to mean both directions of access.

My algorithm for finding a good name includes brainstorming a description of the two concepts and than seeing which keywords to use for the final name.

On the one hand, we have "classes that our class references" (i.e., the referencedInstances in this code), and on the other we have the "classes that reference our class".

Maybe we can separate the terms to have "ReferencedInstances(CaDETClass referencingClass)" and "InstancesThatReference(CaDETClass referencedClass)". Revisit the terminology in your functions to see if every method talks about the correct "reference".

DataSetExplorer/Core/DataSets/Model/RelatedInstance.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace DataSetExplorer.Core.DataSets.Model
+{
+    public class RelatedInstance
+    {
+        public int Id { get; private set; }
+        public string CodeSnippetId { get; private set; }
+        public string Link { get; private set; }
+        public RelationType RelationType { get; private set; }
+        public int CouplingCounter { get; private set; }
```
Luburic, line 9: CouplingStrength is a better term.

Luburic (changes requested):
A few minor things to improve.

Luburic (approved):
Looks good.

PR #101, opened 2022-01-31
Related instances: accessed accessors and fields

simsimkic:


CodeModel/Serialization/ClassCohesionGraphExporter.cs
```diff
@@ -42,7 +42,7 @@ private static ISet<string> GetAllMethodsFieldsAndAccessors(CaDETMember method)
             var retVal = new HashSet<string>();
             retVal.UnionWith(method.GetAccessedOwnFields().Select(f => f.Name));
             retVal.UnionWith(method.GetAccessedOwnAccessors().Select(a => a.Name));
-            retVal.UnionWith(method.InvokedMethods
+            retVal.UnionWith(method.InvokedMethods.Distinct()
```
Luburic, line 45: How did you check if all the Distincts were put in the right place?

Luburic (commented):
One major question to answer.

CodeModel/Serialization/ClassCohesionGraphExporter.cs
simsimkic: Type of InvokedMethods, AccessedAccessors and AccessedFields is changed from _Set_ to _List_. I added Distinct() to get the same effect (without duplicates). I used the list of references provided by the Visual Studio.

